PR #8, opened 2019-10-26
CSV manipulation utilities

satabin:


satabin:
@ybasket If you can take a minute or two to review my generic implementation, it would be nice.

csv/src/fs2/data/csv/CsvRowDecoder.scala
```diff
@@ -0,0 +1,36 @@
...
+
+/** Describes how a row can be decoded to the given type.
+  */
+trait CsvRowDecoder[T, Header] {
+  def apply(row: CsvRow[Header]): DecoderResult[T]
+}
+
+object CsvRowDecoder {
+
+  implicit def RowDecoderFunctor[Header]: Functor[CsvRowDecoder[*, Header]] =
```
ybasket: For consistency, I suggest naming it `CsvRowDecoderFunctor`.

csv/src/fs2/data/csv/package.scala
```diff
@@ -40,4 +43,11 @@ package object csv {
   def noHeaders[F[_]](implicit F: ApplicativeError[F, Throwable]): Pipe[F, NonEmptyList[String], CsvRow[Nothing]] =
     CsvRowParser.pipe[F, Nothing](false)
 
+  def decode[F[_], R](implicit F: ApplicativeError[F, Throwable], R: RowDecoder[R]): Pipe[F, NonEmptyList[String], R] =
```
ybasket: I wonder whether it would make sense to additionally provide `attempt`-versions like
```scala
def attemptDecode[F[_], R](implicit F: Applicative[F], R: RowDecoder[R]): Pipe[F, NonEmptyList[String], DecoderResult[R]]
```
Of course the implementation is extremely simple, but this would be the place where I as a user of the library would look first and be happy to get a solution without diving into details.

csv/src/fs2/data/csv/RowDecoder.scala
```diff
@@ -0,0 +1,36 @@
...
+
+object RowDecoder {
+
+  implicit object RowDecoderFunctor extends Functor[RowDecoder] {
+    def map[A, B](fa: RowDecoder[A])(f: A => B): RowDecoder[B] =
+      cells => fa(cells).map(f)
+  }
+
+  def apply[T: RowDecoder]: RowDecoder[T] = implicitly[RowDecoder[T]]
+
```
ybasket: Maybe we could add a conversion utility from `CsvRowDecoder` to `RowDecoder` here?

csv/src/fs2/data/csv/CsvRow.scala
```diff
@@ -23,13 +23,115 @@ class CsvRow[Header](val values: NonEmptyList[String], val headers: Option[NonEm
...
+        case (cell, i) =>
+          if (i === idx)
+            f(cell)
+          else
+            cell
+      }, headers)
+
+  /** Modifies the cell content at the given `header` using the function `f`.
+    */
+  def modify(header: Header)(f: String => String): CsvRow[Header] =
```
ybasket: The semantics of these methods depend on all `header` values being unique (as it will only modify the first occurrence). Not sure there is actually a convincing use case for having the same value for several columns which would justify changing the current semantics. But at least I suggest documenting it clearly.

csv/src/fs2/data/csv/CellDecoder.scala
```diff
@@ -0,0 +1,77 @@
...
+      s => Right(x)
+
+    def raiseError[A](e: DecoderError): CellDecoder[A] =
+      s => Left(e)
+
+    def tailRecM[A, B](a: A)(f: A => CellDecoder[Either[A, B]]): CellDecoder[B] = {
+      @tailrec
+      def step(s: String, a: A): DecoderResult[B] =
+        f(a)(s) match {
+          case left @ Left(_)          => left.asInstanceOf[DecoderResult[B]] // avoid creating too many objects
```
ybasket: cats has some nice `withRight` and `withLeft` helpers for exactly these casts.
```suggestion
          case left @ Left(_)          => left.withRight[B] // avoid creating too many objects
```

build.sc
```diff
@@ -5,11 +5,12 @@ import publish._
 
 import $ivy.`com.lihaoyi::mill-contrib-bloop:$MILL_VERSION`
 
-val scala212 = "2.12.10"
+val scala212 = "2.12.8"
```
ybasket: A Scala downgrade? 😳

csv/src/fs2/data/csv/package.scala
satabin: True, I will add it.

csv/src/fs2/data/csv/RowDecoder.scala
satabin: yep, I thought about it but forgot, I will add it as well.

build.sc
satabin: Yes, just for local technical reasons, I shouldn't commit this, thanks.

csv/src/fs2/data/csv/CsvRow.scala
satabin: This is true. I'll comment it.

csv/src/fs2/data/csv/RowDecoder.scala
satabin: Actually, it is only possible to convert in the other direction, as a `CsvRowDecoder` might need headers that a `RowDecoder` doesn't have, I will add the other conversion.

csv/src/fs2/data/csv/CellDecoder.scala
satabin: `withRight` expects `B` to be a super type of whatever right type is for `left`. In this case it is an `Either[A, B]`, so it doesn't work, I need `rightCast`.

PR #10, opened 2019-11-27
Add generic derivation for cell decoders

ybasket:
* derives cell decoders for coproducts where each case is
  * an `object` or 
  * a `case object` or
  * has an implicit `CellDecoder` instance
* allows changing the expected value in CSV for objects by using the `@CsvName` annotation
* derives cell decoders for unary product types (case classes with one field)

See tests for some examples.

### DIFF
--- a/csv/generic/src/fs2/data/csv/generic/CsvName.scala
+++ b/csv/generic/src/fs2/data/csv/generic/CsvName.scala
@@ -1,0 +1,9 @@
+package fs2.data.csv.generic
+
+import scala.annotation.Annotation
+
+/**
+ * Mark an object of sealed trait to have a different representation in CSV than the name of the type.
+ * @param name the name to expect in the CSV
+ */
+case class CsvName(name: String) extends Annotation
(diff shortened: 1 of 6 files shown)


ybasket:
@satabin happy to react to your quality concerns ;)

csv/generic/src/fs2/data/csv/generic/CsvName.scala
```diff
@@ -0,0 +1,9 @@
+package fs2.data.csv.generic
+
+import scala.annotation.Annotation
+
+/**
+ * Mark an object of sealed trait to have a different representation in CSV than the name of the type.
+ * @param name the name to expect in the CSV
+ */
+case class CsvName(name: String) extends Annotation
```
satabin: Wouldn't `CsvValue` be a better fit. With `Name` I first understood _header name_.

ybasket: Agree, `CsvValue` is a better name, so I just changed it.

PR #11, opened 2019-11-29
Auto derivation for csv-generic

ybasket:
Using the same technique as in pureconfig and circe. Once #10 is merged, support for cell decoders can be added as well.

Additionally, I added a small beautification to the file-based tests, relativizing the file paths to make the test output more readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,8 @@ val rows = withh.through(decodeRow[IO, String, Row])
 println(rows.compile.toList.unsafeRunSync())
 ```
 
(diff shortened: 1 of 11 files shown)


satabin:
Sadly enough it looks like I have to accept this 😭

ybasket:
What’s the matter? Not a fan of auto-derivation or unhappy you couldn’t find something to improve?

satabin:
The former.

satabin:
I knew I shouldn’t have merged it :sob: https://travis-ci.com/satabin/fs2-data/builds/139190115

PR #13, opened 2019-12-16
Allow usage of CsvName to rename fields in CsvRowDecoder derivation

ybasket:
For cases where your header name in the CSV file differs from the field name in your case class. See test for an example.


ybasket:
Codacy has a false alarm on the access of head on non-empty `HList`…looks like Codacy is not up to shapeless standards. I suggest ignoring.

PR #15, opened 2020-02-14
Generalize header parsing

ybasket:
Let ParseableHeader return errors if parsing fails so that more interesting header parsers can be written and parsing can fail early if headers are not as expected. Also make CsvRow a case class to allow for pattern matching.

Use case example: CSV with translations, first column being the key, the other columns being the languages identified by the headers. As it doesn't make sense to parse the file when the languages are invalid, it's nice to be able to fail on header parsing already instead of having to revalidate the headers on every row.

### DIFF
--- a/csv/src/fs2/data/csv/CsvRow.scala
+++ b/csv/src/fs2/data/csv/CsvRow.scala
@@ -18,10 +18,7 @@ package fs2.data.csv
 import cats.data._
 import cats.implicits._
 
-case class CsvRow[Header](values: NonEmptyList[String], header: Option[Header]) {
-
-  private def byHeader[A](implicit ev: Header =:= NonEmptyList[A]): Option[Map[A, String]] =
-    header.map(_.toList.zip(values.toList).toMap)
+case class CsvRow[Header] private[csv] (values: NonEmptyList[String], header: Option[Header]) {
 
   /** Number of cells in the row. */
   def size: Int = values.size
@@ -47,18 +44,6 @@ case class CsvRow[Header](values: NonEmptyList[String], header: Option[Header]) 
             cell
       }, header)
 
-  /** Modifies the cell content at the given `header` using the function `f`.
-    *
-    * **Note:** Only the first occurrence of the values with the given header
-    * will be modified. It shouldn't be a problem in the general case as headers
-    * should not be duplicated.
-    */
-  def modify[A](head: A)(f: String => String)(implicit ev: Header =:= NonEmptyList[A]): CsvRow[Header] =
-    header match {
-      case Some(header) => modify(header.toList.indexOf(head))(f)
-      case None          => this
-    }
-
   /** Returns the row with the cell at `idx` modifed to `value`. */
   def updated(idx: Int, value: String): CsvRow[Header] =
     if (idx < 0 || idx >= values.size)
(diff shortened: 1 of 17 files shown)


ybasket:
Thinking again about the use case and others where headers play a significant role, I decided to change header parsing from being always by column to the more general way of parsing the whole header row at once. That allows for more validation (for example, check that all expected columns are present, no duplicate column names) and for more interesting types (like pre-building a parse function based on the header). To keep the impact small and still support the previous by column approach, I added type aliases and extension methods which keep many use cases source compatible.

It's definitely a debatable change (which might even pave the way to further changes which for example, could make the `Option[Header]` in `CsvRow` easier to deal with), so discussion is very welcome :)

ybasket:
Codacy only complains about the `null` literal as parent exception for `HeaderError`, can be safely ignored.

csv/src/fs2/data/csv/CsvRow.scala
```diff
@@ -135,16 +65,94 @@ object CsvRow {
...
   }
 
-  def fromNelHeaders[Header](nel: NonEmptyList[(Header, String)]): CsvRow[Header] = {
+  def fromNelHeaders[HeadElem](nel: NonEmptyList[(HeadElem, String)]): CsvNelRow[HeadElem] = {
     val (hs, vs) = nel.toList.unzip
-    new CsvRow(NonEmptyList.fromListUnsafe(vs), NonEmptyList.fromList(hs))
+    CsvRow(NonEmptyList.fromListUnsafe(vs), NonEmptyList.fromList(hs))
+  }
+
+  implicit class CsvRowNelOps[HeadElem](row: CsvNelRow[HeadElem]) {
```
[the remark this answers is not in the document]
ybasket: How would you it then? Note that it is an extension for `CsvRow` objects whose headers are `Nel`-based (= `CsvNelRow`), as these don't work on all header types.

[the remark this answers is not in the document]
satabin: Hmm indeed, I missed that part. Then I think I need to read everything through again, I must have missed something here.

csv/src/fs2/data/csv/package.scala
```diff
@@ -28,17 +29,45 @@ package object csv {
...
   def rows[F[_]](separator: Char = ',')(
       implicit F: ApplicativeError[F, Throwable]): Pipe[F, Char, NonEmptyList[String]] =
     RowParser.pipe[F](separator)
 
+  /** Transforms a stream of strings into a stream of CSV rows. Helper for streams which are not initially char-based,
+   *  prefer to use [[rows]] otherwise.
+   */
+  def rowsFromStrings[F[_]](separator: Char = ',')(
+    implicit F: ApplicativeError[F, Throwable]): Pipe[F, String, NonEmptyList[String]] =
+    _.flatMap(s => Stream.chunk(Chunk.charBuffer(JCharBuffer.wrap(s)))) through RowParser.pipe[F](separator)
```
satabin: Can you use the dot syntax please? Not a big fan of the dotless application :)

satabin: Also is it like really way more efficient to use NIO than just `Stream.emits(s)`? If the performance gain is marginal, then I would prefer not having a `CharBuffer`.

csv/src/fs2/data/csv/package.scala
```diff
@@ -16,8 +16,9 @@
 package fs2
 package data
 
-import csv.internals._
+import java.nio.{CharBuffer => JCharBuffer}
 
+import csv.internals._
```
satabin: I like having my personal relative imports at the beginning and the Java ones at the end (yes, I'm that kind of guy :)).

satabin:
Globally speaking I still miss to see the use case for a non-list header type. It adds some complexity to the types and operation resolution. And I can't seem to come up with a use case for it right now that would convince me totally. If you have such a use case, I would be really happy to hear about it :) Maybe even put it in the README, so that it's clear why we add this complexity.

ybasket:
> Globally speaking I still miss to see the use case for a non-list header type. It adds some complexity to the types and operation resolution. And I can't seem to come up with a use case for it right now that would convince me totally. If you have such a use case, I would be really happy to hear about it :) Maybe even put it in the README, so that it's clear why we add this complexity.

As I wrote above, it is debatable change as it comes with some overhead/complexity. The following is a simplified version of an example I encountered at work:
```scala
sealed trait Language
// instances...
case class I18nValue(translate: Language => String)

implicit val i18nRowDecoder: CsvRowDecoder[(String, I18nValue), String] = (row: CsvRow[String]) => {
    for {
      headers <- row.headers.toRight(new DecoderError(s"CSV file must have headers"))
      langs <- headers.tail
        .traverse { lang =>
          Language
            .withNameInsensitiveOption(lang)
            .toRight(new DecoderError(s"""Unknown supported language "$lang""""))
        }
      i18n <- I18nValue.buildFromAllLanguages(langs zip row.values.tail) // fails if not all languages are present
    } yield row.values.head -> i18n
  }
```

The code would be way cleaner if it was possible to fail processing when parsing the headers already (not all languages present, invalid languages present) and not while handling the rows themselves (which implies the overhead of checking language exhaustivity for each row again). 

Similar early validation use cases I could see when importing CSV files into a RDBMS, by the table structure you can reject CSV files when parsing the headers, but you don't want to check again on every row you insert.

Of course, in case of an error, the stream fails on the first row, but for correct files (which should be the dominant case) it adds overhead not be able to fail header parsing. Plus the necessary workaround makes it harder to deal with bad rows as `attemptDecode` will alway return a `Left` if the headers aren't valid.

Happy to fix all the comments you left in case this convinces you.

PR #16, opened 2020-02-27
Improve handling of headers

ybasket:
Supersedes #15, see discussion there for motivating use cases.

Includes:
* Change `ParseableHeader` to parse all header columns at once to allow for more elaborated validation
* Add `Row` class to represent csv rows without headers properly
* Make `CsvRow` a `case class` extending `Row` and make headers non-optional
* Introduce `skipHeaders` and `withHeaders` pipes to cover all combinations of headers being present in the file and in the desired output


csv/src/fs2/data/csv/ParseableHeader.scala
```diff
@@ -15,12 +15,16 @@
...
+import cats.implicits._
+import cats.{Functor, SemigroupK}
+
+/** A type class describing what it means to be a parseable
   * CSV header.
   */
 trait ParseableHeader[Header] {
 
-  def parse(name: String): Header
+  def apply(names: NonEmptyList[String]): HeaderResult[Header]
```
satabin: Why not use the `F` of the stream it will be used with?

ybasket: For representing errors? Consistency with `CsvRowDecoder`, `RowDecoder` and `CellDecoder` and more narrow error type. 

If we were to go with something like `def apply[F[_]](names: NonEmptyList[String])(implicit me: MonadError[F, Throwable]): F[NonEmptyList[Header]]`, other exception types could be thrown, if we use a `MonadError[F, HeaderError]`, it gets harder to use as there are usually no implicit instances available for that. Similar for `ApplicativeError`, but then you can't even `flatMap`. Might still be an interesting encoding if all decoder types were using it and some inductive instances for the necessary `MonadError` were defined in the project.

Making the effect type a parameter of `ParseableHeader` would allow for launching an array of missiles concurrently, I don't want header parsing to do that.

satabin: I'm definitely not entirely happy with what I created here :( Ok, then.

csv/src/fs2/data/csv/internals/CsvRowParser.scala
```diff
@@ -19,23 +19,22 @@ package csv
...
-        (initialized, Some(new CsvRow[Header](fields, headers)))
-    }.map(_._2).unNone
+  def pipe[F[_], Header](implicit F: ApplicativeError[F, Throwable],
+                         Header: ParseableHeader[Header]): Pipe[F, NonEmptyList[String], CsvRow[Header]] =
+    _.pull.uncons1.flatMap {
+      case Some((firstRow, tail)) =>
+        Header(firstRow) match {
+          case Left(error) => Pull.raiseError[F](error)
+          case Right(headers) if headers.length =!= firstRow.length =>
+            Pull.raiseError[F](new HeaderError("The count of headers must match the count of columns"))
```
satabin: Potentially add the sizes in the message, I think it can be useful when this happens, to debug

csv/src/fs2/data/csv/package.scala
```diff
@@ -35,18 +36,25 @@ package object csv {
...
-  def decode[F[_], R](implicit F: ApplicativeError[F, Throwable], R: RowDecoder[R]): Pipe[F, NonEmptyList[String], R] =
-    _.evalMap(R(_).liftTo[F])
+  /** Transforms a stream of raw CSV rows into rows. */
+  def noHeaders[F[_]]: Pipe[F, NonEmptyList[String], Row] =
+    _.map(new Row(_))
 
-  def attemptDecode[F[_], R](implicit F: Applicative[F],
-                             R: RowDecoder[R]): Pipe[F, NonEmptyList[String], DecoderResult[R]] =
-    _.map(R(_))
+  /** Transforms a stream of raw CSV rows into rows, skipping the first line to ignore the headers. */
```
satabin: First "line" is confusing in the context of CSV, I prefer to use "row", as it might span several lines.

satabin:
2 minor comments, otherwise LGTM.

ybasket:
> 2 minor comments, otherwise LGTM.

Both addressed in afe6c82 :)

PR #17, opened 2020-03-07
Add support for JSON merge patches

satabin:


satabin:
@ybasket just to be sure, can you please have a quick look at this one?

json/test/src/fs2/data/json/mergepatch/JsonMergePatchTest.scala
```diff
@@ -0,0 +1,183 @@
...
+import org.scalatest.matchers.should.Matchers
+
+class JsonMergePatchTest extends AnyFlatSpec with Matchers {
+
+  val valuePatch = JsonMergePatch.Value(Json.fromInt(5))
+  val nullPatch = JsonMergePatch.Value(Json.Null)
+  val objectPatch = JsonMergePatch.Object(
+    Map("key1" -> Json.fromInt(3),
+        "key2" -> Json.obj("nested" -> Json.fromString("s"), "new" -> Json.True, "deleted" -> Json.Null)))
+  val objectPatchTokens = List(
```
ybasket: Maybe something slightly more readable? Like
```scala
val objectPatchTokens = CirceTokenizer.tokenize(
    Json.obj(
      "key1" := 3,
      "key2" := Json.obj(
        "nested" := "s",
        "new" := true,
        "deleted" := Json.Null
      )
    )
  ).toList
```

json/src/fs2/data/json/internal/package.scala
```diff
@@ -17,10 +17,80 @@ package fs2
...
 
-  private[internals] type Result[F[_], In, Out] = Option[(Chunk[In], Int, Stream[F, In], Out)]
+  private[json] type Result[F[_], In, Out] = Option[(Chunk[In], Int, Stream[F, In], Out)]
+
+  private[json] def skipValue[F[_]](chunk: Chunk[Token],
+                                    idx: Int,
+                                    rest: Stream[F, Token],
+                                    depth: Int,
+                                    chunkAcc: List[Token])(
+      implicit F: ApplicativeError[F, Throwable]): Pull[F, Token, Result[F, Token, List[Token]]] =
```
ybasket: I wonder (in general for fs2-data) whether it would make sense to use `RaiseThrowable`…it would allow describing the streams in terms of `Fallible` which makes it a bit more general and detached from the call site. Not sure it's worth it though, what do you think?

README.md
```diff
@@ -117,6 +120,32 @@ val transformed = stream.through(transform[IO, Json](selector, json => Json.obj(
 println(transformed.through(values[IO, Json]).compile.toList.unsafeRunSync())
 ```
 
+### Patches
+
+The `fs2-data-json-diffson` module provides some integration with [diffson][diffson].
+It allows for patching a Json stream as it is read to emit the patched value downstream.
```
ybasket: As it a rather advanced thing, maybe add a use case example (doesn't have to be code, just one or two sentences why this is a cool thing to do)?

json/test/src/fs2/data/json/mergepatch/JsonMergePatchTest.scala
satabin: I can even use `parse` :)

json/src/fs2/data/json/internal/package.scala
satabin: Maybe, I can investigate, but it definitely belongs to another PR.

json/test/src/fs2/data/json/mergepatch/JsonMergePatchTest.scala
satabin: After trying, actually nope it doesn't work as the circe object field map has arbitrary order for keys, whereas the stream tokens are ordered. So sadly, I can't make it more readable.

README.md
satabin: I added a note. Thanks for the suggestions. If you have any other idea, then I'd be glad to hear them.

PR #18, opened 2020-03-11
Add basic JMH benchmark

ybasket:
Enables benchmarking using JMH by setting up the necessary build config and adding a simple csv benchmark with corresponding baseline. To be extended further.


benchmarks/src/fs2/data/benchmarks/CsvParserBenchmarks.scala
```diff
@@ -0,0 +1,42 @@
...
+@Fork(value = 1)
+@Warmup(iterations = 3, time = 2)
+@Measurement(iterations = 5, time = 2)
+class CsvParserBenchmarks {
+
+  var csvContent: List[List[Char]] = _
+
+  @Setup
+  def readCsv(): Unit = csvContent = {
+    File("../../../../../benchmarks/resources/benchmark.csv").contentAsString.grouped(4096).toList.map(_.toList)
```
satabin: Why do you read the entire file into memory? To avoid measuring I/O in the benchmark?

ybasket: Exactly. The idea behind this first, rather basic benchmark is to reveal performance regressions when changing something on the parser, not to test a real life use case. So avoiding external influence and having a baseline benchmark to get an approximation of the time really spent on parsing.

satabin: Ok perfect. LGTM

PR #22, opened 2020-05-05
Use `RaiseThrowable` instead of `ApplicativeError`

satabin:


satabin:
Then, let's not wait any further. Do you want it to be your first merge? ;)

PR #23, opened 2020-05-11
Add Json selector interpolator

satabin:
This is done in another module so that the core Json library only
depends on fs2, and doesn't bring extra dependencies to user who do not
want it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ A set of streaming data parsers based on [fs2][fs2].
 Following modules are available:
   - `fs2-data-json`: [![Maven Central](https://img.shields.io/maven-central/v/org.gnieh/fs2-data-json_2.13.svg)](https://mvnrepository.com/artifact/org.gnieh/fs2-data-json_2.13) A JSON parser and manipulation library
   - `fs2-data-json-circe`: [![Maven Central](https://img.shields.io/maven-central/v/org.gnieh/fs2-data-json-circe_2.13.svg)](https://mvnrepository.com/artifact/org.gnieh/fs2-data-json-circe_2.13) [circe][circe] support for parsed JSON.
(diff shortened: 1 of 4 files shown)


json/contextual/src/fs2/data/json/SelectorInterpolator.scala
```diff
@@ -0,0 +1,35 @@
...
+import cats.implicits._
+
+import contextual._
+
+import scala.util.{Failure, Success, Try}
+
+object SelectorInterpolator extends Verifier[Selector] {
+
+  def check(string: String): Either[(Int, String),Selector] =
+    new SelectorParser[Try](string).parse() match {
```
ybasket: Why `Try` if you're returning `Either`? `Either` and then `leftMap` or `adaptError` would seem more idiomatic to me...

README.md
```diff
@@ -6,6 +6,7 @@ A set of streaming data parsers based on [fs2][fs2].
 Following modules are available:
   - `fs2-data-json`: [![Maven Central](https://img.shields.io/maven-central/v/org.gnieh/fs2-data-json_2.13.svg)](https://mvnrepository.com/artifact/org.gnieh/fs2-data-json_2.13) A JSON parser and manipulation library
   - `fs2-data-json-circe`: [![Maven Central](https://img.shields.io/maven-central/v/org.gnieh/fs2-data-json-circe_2.13.svg)](https://mvnrepository.com/artifact/org.gnieh/fs2-data-json-circe_2.13) [circe][circe] support for parsed JSON.
+  - `fs2-data-json-contextual`: [![Maven Central](https://img.shields.io/maven-central/v/org.gnieh/fs2-data-json-contextual_2.13.svg)](https://mvnrepository.com/artifact/org.gnieh/fs2-data-json-contextual_2.13) [contextual][contextual] support for statically checked JSON interpolators.
```
ybasket, line 9: I would suggest to use a name which indicates what the user can do with the module rather than what dependencies it has, that's implementation detail to me. So maybe `-interpolators` or maybe the almighty `-utils`?

json/contextual/src/fs2/data/json/SelectorInterpolator.scala
satabin: Yeah I notice before seeing your comment and pushed a version with `Either`.

README.md
satabin: Why not.

PR #25, opened 2020-05-16
Add website with documentation

satabin:


documentation/docs/json/index.md
```diff
@@ -0,0 +1,100 @@
...
+stream.compile.toList.unsafeRunSync()
+```
+
+The pipe validates the JSON structure while parsing. It reads all the json values in the input stream and emits tokens as they are available.
+
+### Selectors
+
+Selectors can be used to select a subset of a JSON token stream.
+
+For instance, to select and enumerate elements that are in the `field3` array, you can create this selector. Only the tokens describing the values in `field3` will be emitted as a result.
```
ybasket, line 46: I would mention and link the `interpolators` module here, people might not find it otherwise – who reads full docs ;)

satabin: Right, I added a paragraph.

PR #26, opened 2020-05-18
Make CSV decoders more newcomer friendly

satabin:
- Implement the common mapping operators directly on the decoder
   classes, so that it is easier to discover in IDEs.
 - Add a comment in decoder scaladoc about the cats import to get all
   the operators.
 - Add a `CellDecoder` code snippet in the documentation to desmonstrate
   usage of `emap`.
 - Unify all decoder operators.

### DIFF
--- a/csv/src/fs2/data/csv/CsvRowDecoder.scala
+++ b/csv/src/fs2/data/csv/CsvRowDecoder.scala
@@ -21,17 +21,88 @@ import cats.implicits._
 import scala.annotation.tailrec
 
 /** Describes how a row can be decoded to the given type.
+  *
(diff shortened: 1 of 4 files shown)


csv/src/fs2/data/csv/CsvRowDecoder.scala
```diff
@@ -55,6 +125,11 @@ object CsvRowDecoder extends ExportedCsvRowDecoders {
       }
     }
 
+  implicit def CsvRowDecoderSemigroupK[Header]: SemigroupK[CsvRowDecoder[*, Header]] =
```
ybasket, line 128: Minor consistency thing: Maybe combine the instance of `MonadError` and `SemigroupK` into one as for the other decoder types?

PR #27, opened 2020-05-19
Add more JSON utilities

satabin:
In order to make the full roundtrip, this adds:
 - pretty/compact printing of Json streams
 - tokenizing ASTs

### DIFF
--- a/json/src/fs2/data/json/internal/Renderer.scala
+++ b/json/src/fs2/data/json/internal/Renderer.scala
@@ -1,0 +1,190 @@
+/*
+ * Copyright 2020 Lucas Satabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package fs2
+package data
+package json
+package internals
+
+import scala.annotation.{tailrec, switch}
+
+private[json] class Renderer(pretty: Boolean, resetOnChunk: Boolean, indent: String)
+    extends Collector.Builder[Token, String] {
+
+  import Renderer._
+
+  private val builder = new StringBuilder
+
+  private var level = 0
+
+  private var newline = false
+
+  private var comma = false
+
+  private def separator(comma: Boolean): Unit = {
+    if (comma)
+      builder.append(',')
+  }
+
+  private def indentation(newline: Boolean, level: Int): Unit =
+    if (newline) {
+      builder.append('\n')
+      builder.append(indent * level)
+    }
+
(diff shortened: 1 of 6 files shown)


json/src/fs2/data/json/internal/Renderer.scala
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  def result: String = builder.result()
+
+}
+
+private[json] object Renderer {
+
+  def pipe[F[_]](pretty: Boolean, indent: String): Pipe[F, Token, String] = {
+    val builder = new Renderer(pretty, true, indent)
```
ybasket, line 167: I have the suspicion this isn't safe 🤔

If you acquire this pipe once and then use it in concurrently for several streams, I feel like these could interfere in the local state of the `Renderer`. To be sure, the renderer should probably be acquired effectfully or at least suspended:

```scala
in => Stream.suspend(Stream.pure(new Renderer(pretty, true, indent))).flatMap { renderer =>
  in.mapChunks { chunk =>
       renderer += chunk
       Chunk.singleton(renderer.result)
     }
}
```

satabin: You are correct, mister, I'll fix this.

satabin: I wonder if I can't just
```scala
    { in =>
      val renderer = new Renderer(pretty, true, indent)
      in.mapChunks { chunk =>
       renderer += chunk
       Chunk.singleton(renderer.result)
      }
    }
```

satabin: Nevermind, of course I can’t…

json/src/fs2/data/json/internal/Renderer.scala
```diff
@@ -0,0 +1,188 @@
...
+
+  private def indentation(newline: Boolean, level: Int): Unit =
+    if (newline) {
+      builder.append('\n')
+      builder.append(indent * level)
+    }
+
+  private def jsonString(s: String, key: Boolean): Unit = {
+    @tailrec
+    def loop(idx: Int): Unit =
```
satabin, line 49: @ybasket I also fixed escaping, but for the sake of "you never know what you do on a Saturday morning, so better safe than sorry", can you quickly check this new version?

json/src/fs2/data/json/package.scala
```diff
@@ -58,6 +58,66 @@ package object json {
...
+
+    /** Renders a pretty-printed representation of the token stream with the given
+      * indentation size.
+      *
+      * Chunks can be concatenated to render all values in the stream,
+      * separated by new lines.
+      *
+      * You can use this to write the Json stream to a file.
+      */
+    def pretty[F[_]](indent: Int = 2): Pipe[F, Token, String] =
```
ybasket, line 89: Both for here and the collector: Maybe let's not choose a side on the spaces vs tabs war and allow to specify the indentation string as another parameter with default? Maybe also the newline? No need to reproduce the full power of circe's pretty printer or similar, but these options seem very straight-forward to add…

json/src/fs2/data/json/internal/Renderer.scala
ybasket: It's also Saturday morning for me ;)

But looks all reasonable!

json/src/fs2/data/json/package.scala
satabin: :sob: Ok…

PR #28, opened 2020-05-26
Spelling fix

randbw:


csv/src/fs2/data/csv/CellDecoder.scala
```diff
@@ -49,7 +49,7 @@ trait CellDecoder[T] {
     s => apply(s).map(f)
 
   /**
-    * Map the parsed value to a new decoder, which in turn will be applie toString
+    * Map the parsed value to a new decoder, which in turn will be applied toString
```
ybasket, line 52: Hey, thank you for the PR! Would you mind fixing the `toString` to `to` as well so that we have proper English sentence here?

randbw: done 👯

PR #31, opened 2020-06-04
Add abstraction to iterate over characters

satabin:


satabin:
poke @ybasket

benchmarks/src/fs2/data/benchmarks/CsvParserBenchmarks.scala
```diff
@@ -16,19 +16,19 @@ import cats.effect.SyncIO
...
   }
 
-  def csvStream: Stream[SyncIO, Char] = Stream.emits(csvContent).flatMap(Stream.emits).covary[SyncIO]
+  def csvStream: Stream[SyncIO, String] = Stream.emits(csvContent).covary[SyncIO]
 
   @Benchmark
   def parseRows(): Unit = {
     csvStream
-      .through(fs2.data.csv.rows[SyncIO](','))
+      .through(fs2.data.csv.rows[SyncIO, String](','))
```
ybasket, line 31: Have you run this benchmark to see the difference before and after this PR? I love the new abstraction, but given that it so critical for the performance of the whole library, IMHO we should run (and keep) some benchmarks to prove this is worth it. Maybe a simple benchmark comparing dumping `Char`s into a `Blackhole` from the old `Stream.flatMap(Stream.emits(_))` vs the new `CharLikeChunks` on a stream of `String`s?

text/src/fs2/data/text/CharLikeChunks.scala
```diff
@@ -0,0 +1,188 @@
...
+trait CharLikeChunks[F[_], In] {
+
+  /** The context used to pull chunks and iterate over characters.
+    * Implementations might decide to have immutable or mutable contexts
+    * depending on performance constraints that are desired.
+    */
+  type Context
+
+  /** Creates a context out of a stream. */
+  def create(s: Stream[F, In]): Context
```
ybasket, line 37: Maybe it would be better to suspend the creation in an effect? The current way works because every implementation does a `Stream.suspend` and it's all internal, but generally speaking, it shouldn't be the caller's responsibility to take care of that.

benchmarks/src/fs2/data/benchmarks/CsvParserBenchmarks.scala
satabin: Yes I did, with the code you added, here are the results:

```
# master

Benchmark                              Mode  Cnt      Score     Error  Units
CsvParserBenchmarks.parseRows          avgt    5  11047,991 ± 118,289  us/op
CsvParserBenchmarks.parseRowsBaseline  avgt    5   6290,640 ±  28,784  us/op

# charlikechunks

Benchmark                              Mode  Cnt     Score    Error  Units
CsvParserBenchmarks.parseRows          avgt    5  4239,121 ± 38,023  us/op
CsvParserBenchmarks.parseRowsBaseline  avgt    5     2,877 ±  0,025  us/op
```

The difference is huge, but since I also refactored the CSV parser to emit non-1 chunks, it might be explained by a combination of both things. I will add benchmarks for JSON and CSV, since they didn't have this big refactoring

text/src/fs2/data/text/package.scala
```diff
@@ -15,10 +15,25 @@
...
 package data
-package xml
 
-package object internals {
+import java.nio.charset.Charset
 
-  private[internals] type Result[F[_], O] = (Context[F], O)
+package object text {
+
+  /** Import this if your byte stream is encoded in ISO-8859-15 (aka latin1) */
```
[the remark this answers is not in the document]
satabin: I will fix this and add both latin1 and latin2.

I thought about making the constructor available to all for any charset, but it is easy to create something that won't work by mistake if you pass a charset that as multi-byte encoding, even with the runtime check. I prefer exposing predefined ones.

text/src/fs2/data/text/CharLikeChunks.scala
satabin: I don't want to force the `Sync` on `F`.

satabin: And even if it is an `F` there is no way you can enforce `F.pure` won't be used in this. I prefer doing it for sure at call site. Because here, this really is necessary because of how the call site uses it, so I prefer keeping it that way.

text/src/fs2/data/text/package.scala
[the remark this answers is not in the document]
ybasket: An interpolator would move this to compile-time, should hence be safe. But can also always be added later on.

benchmarks/src/fs2/data/benchmarks/CsvParserBenchmarks.scala
ybasket: That's roughly a ~2.5 factor 😳 IMHO too big to ignore, so excited on how it performs on XML and JSON.

text/src/fs2/data/text/CharLikeChunks.scala
ybasket: Well, as always in Scala you'll find a way to shoot yourself in the foot if you want to – types can only give guidance. But I see your point about the `Sync` constraint, so maybe leave as-is and just add a warning in the Scaladoc about this call contract?

satabin: Yes, I am writing contribution guidelines currently, including what to be aware of when adding new parsers. This will definitely be part of this as well.

text/src/fs2/data/text/package.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import java.nio.charset.Charset
+
+package object text {
+
+  /** Import this if your byte stream is encoded in ISO-8859-1 (aka latin1) */
+  object latin1 {
+
+    implicit def byteStreamCharLike[F[_]]: CharLikeChunks[F, Byte] =
+      new CharLikeSingleByteChunks[F](Charset.forName("ISO-8859-1"))
```
ybasket, line 27: There are constants for ISO-8859-1 and ASCII in `StandardCharsets`, might want to use them.

PR #37, opened 2020-06-15
Add mandatory field selector

### LINKED ISSUE #30: Add ability to signal selection failure for streaming nested json extraction
Recently, we hit an issue where our code that used `.key.[]` emitted no elements in its stream, because `.key` was incorrect - no such key was in the json.

I'd like to have a mode for filters that can error out if the key being filtered isn't present

satabin:
This selector fails the stream if a selected field is not present in the
object the selector is applied to. It can be useful when you know your
data must contain the field and want to raise an error if not (e.g. to
detect typos).

fixes #30

### DIFF
--- a/documentation/docs/json/index.md
+++ b/documentation/docs/json/index.md
@@ -33,8 +33,8 @@ val input = """{
               |  "field3": []
               |}""".stripMargin
 
-val stream = Stream.emits(input).through(tokens[IO])
-stream.compile.toList.unsafeRunSync()
+val stream = Stream.emits(input).through(tokens[Fallible])
+stream.compile.toList
 ```
 
 The pipe validates the JSON structure while parsing. It reads all the json values in the input stream and emits tokens as they are available.
@@ -46,9 +46,13 @@ Selectors can be used to select a subset of a JSON token stream.
 For instance, to select and enumerate elements that are in the `field3` array, you can create this selector. Only the tokens describing the values in `field3` will be emitted as a result.
 
 ```scala mdoc
-val selector = ".field3.[]".parseSelector[IO].unsafeRunSync()
+import cats.implicits._
+
+type ThrowableEither[T] = Either[Throwable, T]
+
+val selector = ".field3.[]".parseSelector[ThrowableEither].toTry.get
 val filtered = stream.through(filter(selector))
-filtered.compile.toList.unsafeRunSync()
+filtered.compile.toList
 ```
 
 The `parseSelector` method implicitly comes from the `import fs2.data.json._` and wraps the result in anything that has an [`MonadError` with error type `Throwable`][monad-error] to catch potential parsing errors. If you prefer not to have this wrapping and don't mind an extra dependency, you can have a look at [the interpolator][interpolator-doc].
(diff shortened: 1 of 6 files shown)


json/src/fs2/data/json/JsonException.scala
```diff
@@ -15,4 +15,4 @@
  */
 package fs2.data.json
 
-class JsonException(msg: String) extends Exception(msg)
+case class JsonException(msg: String) extends Exception(msg)
```
ybasket, line 18: Could be handy to know which fields were missing without having to match the message, wdyt?

documentation/docs/json/index.md
```diff
@@ -94,7 +111,7 @@ If you provide an implicit [`Tokenizer[Json]`][tokenizer-api], which describes h
 
 ```scala
 implicit tokenizer: Tokenizer[SomeJsonType] = ...
-val transformed = stream.through(transform[IO, Json](selector, json => SomeJsonObject("test" -> json)))
+val transformed = stream.through(transform[Fallible, Json](selector, json => SomeJsonObject("test" -> json)))
 ```
 For concrete examples of provided `Builder`s and `Tokeizer`s, please refer to [the JSON library binding modules documentation][json-lib-doc]
```
ybasket, line 116:
```suggestion
For concrete examples of provided `Builder`s and `Tokenizer`s, please refer to [the JSON library binding modules documentation][json-lib-doc]
```

json/src/fs2/data/json/SelectorParser.scala
```diff
@@ -86,15 +86,21 @@ class SelectorParser[F[_]](val input: String)(implicit F: MonadError[F, Throwabl
...
       case None =>
         F.raiseError(new JsonSelectorException("unexpected end of input", idx))
     }
 
-  private def parseSeveralNames(idx: Int, names: Seq[String]): F[(Boolean => Selector, Int)] =
+  private def parseMandatory(idx: Int, names: List[String]): F[(Boolean => Selector, Int)] =
+    next(idx, peek = true).flatMap {
+      case Some((Bang(_), idx)) => F.pure((Selector.NameSelector(names, _, true), idx + 1))
+      case _ => F.pure((Selector.NameSelector(names, _, false), idx))
+    }
```
ybasket, line 101: No need for `flatMap` as both cases call `F.pure`:
```suggestion
    next(idx, peek = true).map {
      case Some((Bang(_), idx)) => (Selector.NameSelector(names, _, true), idx + 1)
      case _ => (Selector.NameSelector(names, _, false), idx)
    }
```

json/src/fs2/data/json/selectors.scala
```diff
@@ -28,16 +28,18 @@ object Selector {
...
+    def apply(name: String, strict: Boolean, mandatory: Boolean) =
+      new NameSelector(NamePredicate.Single(name), strict, mandatory)
+    def apply(names: Seq[String], strict: Boolean, mandatory: Boolean) =
       names match {
-        case Seq(name) => new NameSelector(NamePredicate.Single(name), strict)
-        case Seq()     => new NameSelector(NamePredicate.None, strict)
-        case _         => new NameSelector(NamePredicate.Several(names.toSet), strict)
+        case Seq(name) => new NameSelector(NamePredicate.Single(name), strict, mandatory)
+        case Seq()     => new NameSelector(NamePredicate.None, strict, mandatory)
+        case _         => new NameSelector(NamePredicate.Several(names.toSet), strict, mandatory)
```
ybasket, line 42: Is there a reason why they get parsed to a `List`, but converted to a `Set` here? Maybe parse the names directly to a `Set`?

json/src/fs2/data/json/JsonException.scala
satabin: Could be, I was too lazy too add a specific exception but since you are asking :)

json/src/fs2/data/json/SelectorParser.scala
satabin: Indeed.

json/src/fs2/data/json/selectors.scala
satabin: Probably no good beyond me being used to aggregate with `::`. I will build the set directly since order doesn't matter.

satabin:
> Apart from the test compilation failures which I think are easy to fix, it looks good!

:man_facepalming:

PR #38, opened 2020-06-25
Ensure all values are emitted before raising

satabin:
The idea is to have a stream that emits all valid values up to the first
exception. Discarding the currently accumulating chunk can result in
some data being sometimes emitted or not, depending on the chunk size.
Observable behavior of the stream should not depend on how data is
chunked, so all valid data _so far_ must always be emitted.

### DIFF
--- a/json/src/fs2/data/json/internal/package.scala
+++ b/json/src/fs2/data/json/internal/package.scala
@@ -23,6 +23,9 @@ package object internals {
 
   private[json] type Result[F[_], In, Out] = Option[(Chunk[In], Int, Stream[F, In], Out)]
 
+  private[json] def emitChunk[T](chunkAcc: List[T]) =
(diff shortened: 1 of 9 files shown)


json/src/fs2/data/json/internal/package.scala
```diff
@@ -23,6 +23,9 @@ package object internals {
 
   private[json] type Result[F[_], In, Out] = Option[(Chunk[In], Int, Stream[F, In], Out)]
 
+  private[json] def emitChunk[T](chunkAcc: List[T]) =
```
ybasket, line 26: Not sure that would make any difference, but maybe add `@inline`?

satabin: Might be, this is part of a broader task with benchmarking to see if it does, and the library is not compiled with the optimizer anway.

PR #43, opened 2020-07-13
Add JSON selector DSL

### LINKED ISSUE #39: Add JSON selector DSL
Currently there are three ways to build JSON selectors:
 1. build them using the case class constructors directly;
 2. use the selector parser
 3. use the selector interpolator

They all work pretty well but depending on the context might not be the best suited ones:
 1. not really intuitive to write
 2. parser returns the result in an effect
 3. interpolator adds an extra dependency

A fourth way would be a small DSL, allowing to write the selectors in scala in a more convenient way than using the constructors directly. As an inspiration, one can think about the [circe `ACursor`](https://circe.github.io/circe/api/io/circe/ACursor.html) class.

This DSL can offer a convenient, pure way to build selectors without any extra dependency.

satabin:
Closes #39.


satabin:
@ybasket sorry for the back and forth, now it is ready to be reviewed.

json/src/fs2/data/json/selector/dsl.scala
```diff
@@ -0,0 +1,175 @@
...
+  type Out <: SelectorBuilder[Mandatory, S]
+  def makeMandatory(builder: In): Out
+}
+
+object Mandatoriable {
+  @implicitNotFound(msg = "Only not yet mandatory field selectors can be made mandatory")
+  type Aux[In, Out0, S] = Mandatoriable[In, S] { type Out = Out0 }
+}
+
+object SelectorBuilder {
```
ybasket, line 156: Very clever encoding, I like it a lot!

Just for the readability of the source code, I would push this companion object up, next to the `SelectorBuilder` class.

json/src/fs2/data/json/selector/dsl.scala
```diff
@@ -0,0 +1,175 @@
...
+  implicit def MandatoriableNames[S]: Mandatoriable.Aux[NamesBuilder[Optional, S], NamesBuilder[Mandatory, S], S] =
+    new Mandatoriable[NamesBuilder[Optional, S], S] {
+      type Out = NamesBuilder[Mandatory, S]
+      def makeMandatory(builder: NamesBuilder[Optional, S]): NamesBuilder[Mandatory, S] =
+        builder.copy(mandatory = true)
+    }
+
+}
+
+class NotApplicable private {}
```
ybasket, line 130: Let's add a small comment about the meaning of each of these markers? Makes reading this file a lot easier if you don't have to figure out by yourself what means what and what is the dual of which other marker.

json/src/fs2/data/json/selector/dsl.scala
satabin: Thanks. And true.

json/src/fs2/data/json/selector/dsl.scala
satabin: Yep.

PR #46, opened 2020-07-25
Switch to weaver test framework

satabin:
Weaver works out of the box with cats-effect and allows for writing
tests in a nice way with an `IO`. Moreover, expectations having a
monoid, it is quite elegant to combine test results from a stream using
`foldMonoid`. This is better than running tests with `unsafeRunSync`.


satabin:
> Awesome! I assume the doc build failure is unrelated? Feel free to merge if so

Thanks.

The build failure is due to, I have of clue what. I thought an update of mdoc or mill, and reverted both versions and still get an error locally. I start being tired of these weird intermittent classpath issues and might come back to sbt until mill stabilizes for real.

ybasket:
Looks like now it can't even work as you changed Travis, but no longer the build itself?

Let me know if I can be of help resolving the issue!

satabin:
@ybasket I rebased all branches tu have the sbt build, you can have another look at this one.

ybasket (approved):
Looks good to me. I don't know how it'll interfere with the changes of #63, so I leave the merge to you, time it as you think is best @satabin

satabin:
> Looks good to me. I don't know how it'll interfere with the changes of #63, so I leave the merge to you, time it as you think is best @satabin

I indeed prefer to wait until #63 is merged and rebase this branch on top of it.

PR #49, opened 2020-07-28
Add more Json transformation pipes

### LINKED ISSUE #47: Add more Json Selector Pipes
Currently one can use Json Selectors to only keep some values or transform a Json value at the selected position into another one. Interesting other use cases would be:
 - decide to keep or filter out a value at a given path based on some predicate
 - having an effectful transformation in case it might fail

This can be achieved by adding following `Pipe`s:

```scala
def transformF[F[_], Json](sel: Selector, f: Json => F[Json]): Pipe[F, Token, Token]

def transformOpt[F[_], Json](sel: Selector, f: Json => Option[Json]): Pipe[F, Token, Token]
```

satabin:
One of them allows for filtering values out of the stream, which
includes to be cautious to be sure the emitted stream is still valid. We
have to make sure that the key was not emitted if the value is going to
be dropped. This introduces a way to delay the emission of a key until
we know for sure the value is kept.

Fixes #47

### DIFF
--- a/json/src/fs2/data/json/package.scala
+++ b/json/src/fs2/data/json/package.scala
@@ -51,7 +51,35 @@ package object json {
   def transform[F[_], Json](selector: Selector, f: Json => Json)(implicit F: RaiseThrowable[F],
                                                                  builder: Builder[Json],
                                                                  tokenizer: Tokenizer[Json]): Pipe[F, Token, Token] =
+    TokenSelector.transformPipe[F, Json](selector, f.andThen(Some(_)))
+
+  /** Transforms a stream of token into another one. The transformation function `f` is
+    * called on every selected value from upstream, and the resulting value replaces it.
+    * If the function returns `None`, then the entire value is dropped (and the object key it
+    * is located at, if any).
+    * The rest of the stream is left unchanged.
+    *
+    * This operator locally creates Json AST values using the [[Builder]], and
+    * returns tokens as emitted by the [[Tokenizer]] on the resulting value.
+    */
+  def transformOpt[F[_], Json](selector: Selector, f: Json => Option[Json])(implicit F: RaiseThrowable[F],
+                                                                 builder: Builder[Json],
+                                                                 tokenizer: Tokenizer[Json]): Pipe[F, Token, Token] =
     TokenSelector.transformPipe[F, Json](selector, f)
+
+  /** Transforms a stream of token into another one. The transformation function `f` is
+    * called on every selected value from upstream, and the resulting value replaces it.
+    * The rest of the stream is left unchanged. The operation can fail, in case the returned
+    * `F` is failed at one step.
+    *
+    * This operator locally creates Json AST values using the [[Builder]], and
+    * returns tokens as emitted by the [[Tokenizer]] on the resulting value.
+    */
+  def transformF[F[_], Json](selector: Selector, f: Json => F[Json])(
+      implicit F: RaiseThrowable[F],
+      builder: Builder[Json],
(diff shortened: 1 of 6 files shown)


json/src/fs2/data/json/package.scala
```diff
@@ -51,8 +51,36 @@ package object json {
   def transform[F[_], Json](selector: Selector, f: Json => Json)(implicit F: RaiseThrowable[F],
                                                                  builder: Builder[Json],
                                                                  tokenizer: Tokenizer[Json]): Pipe[F, Token, Token] =
+    TokenSelector.transformPipe[F, Json](selector, f.andThen(Option(_)))
```
ybasket, line 54:
```suggestion
    TokenSelector.transformPipe[F, Json](selector, f.andThen(Some(_)))
```
No need for checking against `null` values, I suppose?

satabin: And why not? This is user code, and user could return whatever, so I think better safe than sorry.

satabin:
> One minor comment, otherwise LGTM!

I'll had documentation before merging, probably tonight.

json/src/fs2/data/json/package.scala
ybasket: I leave being overly smart to frameworks and tend to KISS libraries. There's a clear way of removing values, so I don't see why someone should return `null` here. So why have everybody pay the (small) price?

Side note: Just looking at the code, `null` would actually bubble through to the final `Stream` without exploding, right? That could be used for interesting hacks like counting the amount of removed elements (of course no FP would do it) ;)

Seeing good arguments for both and having only a slight tendency myself, I leave the decision to you.

json/src/fs2/data/json/package.scala
```diff
@@ -51,8 +51,36 @@ package object json {
...
 
+  /** Transforms a stream of token into another one. The transformation function `f` is
+    * called on every selected value from upstream, and the resulting value replaces it.
+    * The rest of the stream is left unchanged. The operation can fail, in case the returned
+    * `F` is failed at one step.
+    *
+    * This operator locally creates Json AST values using the [[Builder]], and
+    * returns tokens as emitted by the [[Tokenizer]] on the resulting value.
+    */
+  def transformF[F[_], Json](selector: Selector, f: Json => F[Json])(
```
ybasket, line 78: Just wondered whether a `transformOptF` would make sense for completeness? Or thinking even more general, something like 
```scala
def flatTransform[F[_], Json](selector: Selector, f: Json => Stream[F, Json])(...)
```

ybasket: Oh, I guess the `flatTransform` could break the JSON structure in some cases…so maybe only `transformOptF`

json/src/fs2/data/json/package.scala
satabin: I don't see the price as big, and as a functional library I would expect no `null` to be emitted. I prefer not allowing hacks based on `null` and extend and fix library if the needs exists. Also to make it clear, I will move the selector based pipes into the `selector` package and deprecate the ones in the `json` package for now (to be removed in 1.0.0). This will allow for other selector things based on another approach without having these selector always in scope. I can change the behavior only in the new package, then.

json/src/fs2/data/json/package.scala
satabin: Yes, I'd like these pipes to emit valid sequence of tokens if the input steam was valid. Returning a stream here cannot guarantee this. I also thought about `transformOptF` and the more I think about it, the more I think I should do for completeness sake, but I wonder whether I want the `F` versions at all ^^

PR #50, opened 2020-08-05
Customize csv quote handling

drobert:
We had a use case recently where we have an unquoted CSV to parse that contains values with quotation marks. An example might be:

```
name,age,description
John Doe,75,"Starts with Quotes" but ends without
```

The current parser assumes a field beginning with a quote is a quoted field, per RFC, but the data here doesn't reflect that and we need all quotes to be treated literally.

This PR allows for a parsing flag to treat quotes either RFC-compliant or literal.

Note: it probably could use a dedicated test. I wanted to put this up first to get some eyes on it, however.

### DIFF
--- a/csv/src/fs2/data/csv/internals/RowParser.scala
+++ b/csv/src/fs2/data/csv/internals/RowParser.scala
@@ -18,6 +18,8 @@ package data
 package csv
 package internals
 
+import cats.data.{State => _, _}
+
 private[csv] object RowParser {
 
   def pipe[F[_]](separator: Char,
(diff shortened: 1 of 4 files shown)


csv/src/fs2/data/csv/internals/RowParser.scala
```diff
@@ -18,11 +18,11 @@ package data
 package csv
 package internals
 
-import cats.data.{State => _, _}
-
 private[csv] object RowParser {
 
-  def pipe[F[_]](separator: Char)(implicit F: RaiseThrowable[F]): Pipe[F, Char, NonEmptyList[String]] = {
+  def pipe[F[_]](separator: Char,
+                 quoteHandling: QuoteHandling = QuoteHandling.Adaptive)(
```
ybasket, line 24: I guess `Adaptive` is a renaming artefact and should be `QuoteHandling.RFCCompliant`?

drobert: This is what happens when I try to refactor between meetings. Fixing these up. But yes, renamed.

satabin:
It looks good at a first sight, maybe a test case and an addition to the doc, that would be awesome.

drobert:
Put up a test

satabin:
Looks amazing, thanks a lot!

PR #52, opened 2020-08-08
Check consistency between row and header length

### LINKED ISSUE #51: Csv decoding should fail when row width does not match header width
Consider an ill-formed file such as this: (spaces around `\t` for readability; assume they aren't there)

```
user-generated-text \t some-number
First line is ok \t 1
Unescaped\t raw tab in here \t 2
```

Note that the last line there has an extra tab character because the source of the file did not properly escape data from user input

Currently what happens is that a `CsvRowDecoder` *coincidentally* fails because for this one schema, the "shifted" column ends up putting `raw tab in here` into the `some-number` header column, which is using a `CellDecoder[Int]`

But if our `CellDecoder` had instead been expecting to receive a string there, I believe it would *pass without error*.

When this condition happens `(row.length != headers.length)`, the entire stream should abort with an error, rather than continuing to emit potentially malformed results. This check would also catch a case where a user-input `\n` in the middle of an unescaped field would cause nonsense lines.

satabin:
When a header row is required by the user, then it is better to check
consistency of following rows, so that we do not emit bogus rows where
length of headers and values do not match. This can lead to hard to spot
and debug problems.

Fixes #51

### DIFF
--- a/csv/generic/src/fs2/data/csv/generic/MapShapedCsvRowEncoder.scala
+++ b/csv/generic/src/fs2/data/csv/generic/MapShapedCsvRowEncoder.scala
@@ -29,7 +29,7 @@ object MapShapedCsvRowEncoder extends LowPrioMapShapedCsvRowEncoderImplicits {
       ev: <:<[Anno, Option[CsvName]],
       witness: Witness.Aux[Key]): WithAnnotations[Wrapped, FieldType[Key, Repr] :: HNil, Anno :: HNil] =
     (row: Repr :: HNil, annotation: Anno :: HNil) =>
-      CsvRow(NonEmptyList.one(Last(row.head)), NonEmptyList.one(annotation.head.fold(witness.value.name)(_.name)))
+      new CsvRow(NonEmptyList.one(Last(row.head)), NonEmptyList.one(annotation.head.fold(witness.value.name)(_.name)))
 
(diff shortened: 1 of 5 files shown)


ybasket:
I think the equal lengths of headers and values is an invariant to `CsvRow`. Hence, I tend to say we should put the check there and not in the parsers, for example by hiding the constructor and providing an `apply` method which forces you to handle violations by returning some `Either[..., CsvRow]` or similar. That way, we're sure we can't accidentally have corrupt instances by forgetting the check in the parser. WDYT?

satabin:
> I think the equal lengths of headers and values is an invariant to `CsvRow`. Hence, I tend to say we should put the check there and not in the parsers, for example by hiding the constructor and providing an `apply` method which forces you to handle violations by returning some `Either[..., CsvRow]` or similar. That way, we're sure we can't accidentally have corrupt instances by forgetting the check in the parser. WDYT?

Good Idea. It will also allow me to use `rethrow` instead of `flatMap` to handle failure.

satabin:
@ybasket Pushed the change. There might be a way to use shapeless in the generic module to statically ensure that calling the constructor is safe, but I am not sure it is worth the hassle, since the two invocations of the `CsvRow` constructor in the module are pretty safe.

PR #54, opened 2020-08-10
Add HeaderSizeError for downstream consumers to match on

Daenyth:
When raising header size mismatches, use this specific error.

It will cause, for example, exception reporting software like Sentry to
correlate these exceptions separately from logical errors like a bad Decoder

This should be backwards compatible since it's a subclass


satabin:
While we are at it, could it make sense to also have two extra fields `actual` and `expected`?

Daenyth:
I thought about that, but it wasn't clear to me that it was useful beyond just knowing *that* it happened. But it's cheap to add, so sure.

PR #59, opened 2020-09-28
Switch to sbt

satabin:
mill is a nice tool but not mature enough and results in some
random instabilities from time to time (basically, whenever version is
upgraded).
sbt has problems on its own, but a better and more richer ecosystem and
support.


satabin:
This time the PR is ready for real! If you can have a quick look at it, it would be great.

cquiroz (approved):
Looks great, I could take  a stab at cross compiling to js

ybasket (approved):
Looks good!

PR #60, opened 2020-09-30
Add Json stream wrappers

### LINKED ISSUE #48: Add Json Stream Wrappers
A use case for fs2-data might be to generate Json data in a streaming manner and not transform parsed data. I this case it can be interesting to have pipes to wrap the produced data into some array/object structure in a safe way (i.e. while ensuring the produced sequence of tokens is valid).

This is based on a [question asked on gitter](https://gitter.im/fs2-data/general?at=5f08777b64ef9d04b2a5034d).

satabin:
Closes #48

### DIFF
--- a/json/src/fs2/data/json/package.scala
+++ b/json/src/fs2/data/json/package.scala
@@ -94,6 +94,45 @@ package object json {
   def tokenize[F[_], Json](implicit tokenizer: Tokenizer[Json]): Pipe[F, Json, Token] =
     _.flatMap(value => Stream.emits(tokenizer.tokenize(value).toList))
 
+  /** A collection of pipes to wrap streams inside objects. */
+  object wrap {
+
+    /** Wraps the stream elements as an array inside an object at the given `at` key.
+      * The object also contains the keys from the `in` map if any.
+      * If `mapFirst` is true, then the map elements are emitted first, then the stream, otherwise the stream is emitted first.
+      *
+      * The resulting token stream is a valid single JSON object stream, iff the original
+      * stream is a valid stream of JSON values.
+      */
+    def asArrayInObject[F[_], Json](
+        at: String,
+        in: Map[String, Json] = Map.empty[String, Json],
+        mapFirst: Boolean = true)(implicit tokenizer: Tokenizer[Json]): Pipe[F, Token, Token] =
+      ObjectWrapper.pipe[F, Json](at, in, mapFirst, false)
+
+    /** Wraps the stream element as a single value inside an object at the given `at` key.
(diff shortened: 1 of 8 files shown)


json/src/fs2/data/json/package.scala
```diff
@@ -94,6 +94,33 @@ package object json {
...
+  object wrap {
+
+    /** Wraps the stream elements as an array inside an object at the given `at` key.
+      * The object also contains the keys from the `in` map if any.
+      * If `mapFirst` is true, then the map elements are emitted first, then the stream, otherwise the stream is emitted first.
+      *
+      * The resulting token stream is a valid single JSON object stream, iff the original
+      * stream is a valid stream of JSON values.
+      */
+    def asArray[F[_], Json](at: String, in: Map[String, Json] = Map.empty[String, Json], mapFirst: Boolean = true)(
```
ybasket, line 107: `asArray` makes me think of `[e1, e2, ...]`, not `{"a":true, "s":[e1, e2, ...]}` and I believe both have their use cases (though the first has a way simpler implementation, a pipe here could be user-friendly). I'm also thinking of deeply nested structures which now you can build by using several of these pipe, but a selector-based method like `inject[F[_]](wrapper: Stream[F, Token], at: Selector, early: Boolean = true): Pipe[F, Token, Token]` might be more user-friendly.

All these can be implemented in future PRs, I just want mention so that you can reconsider naming. For that, I'll approve, but not merge yet, do you as think. Definitely a very nice PR with awesome documentation, really beautiful!

ybasket (approved):

json/src/fs2/data/json/package.scala
satabin: Actually the name is `wrap.asArray(at = "key")` which is quite explicit I think. I miss the SmallTalk syntax for this kind of DSL. I was thinking about adding a `asTopLevelArray` as well, I can do it. Potentially rename `asArray` into something like `asArrayInObject` but I like it less.

I though about the selector and rejected it because it's hard to make it work properly in a streaming fashion and a user friendly way. Think about how you would put the stream into
```json
{
  "a": {
    "b": {
      "c": true
   },
  "d": 1
}
```
at selector `.a.b.c.d` with `early = true`.

Arbitrary depth is just only corner cases, or then putting constraints into the doc and hoping

satabin: And on top of that, selectors allow for `.a.[].b`, how to interpret it? We would need JSON Path instead of selector here (the later are intended to be used to enumerate and select potentially several values, not identifying a unique one). This is something I thought about as well, but I think it's outside of the scope of this PR.

PR #63, opened 2020-10-03
Cross build some projects to scala.js

### LINKED ISSUE #58: Publish for scala.js
I'm using fs2-data-xml for library that I'd like to cross publish to scala.js.
It seems that it should be possible given the dependencies.
I may try myself but I don't know much about using `mill`.
Tests are problematic though as they use files and that would be harder on Scala.js

cquiroz:
This PR adds cross building to js to most modules (Only interpreters can't be converted due to the contextual dependency)

I changed the dir where srcs are located as it is what the scala.js plugin expects.

Another change is that `URL` isn't avilable for js, I changed the `CellEncoder/CellDecder` for `URL` to `URI`. It is possible to restore the `URL` one for jvm only. Let me know if you prefer it

This complicates a bit the build but I'm glad that most modules can be converted.
I've tested running the xml module on the browser already

Fixes #58


cquiroz:
Note that I try to run all tests in both platforms but leaving the ones that need files to run on the JVM only

csv/generic/src/test/scala/fs2/data/csv/generic/AutoDerivationTest.scala
```diff
@@ -40,7 +40,6 @@ class AutoDerivationTest extends AnyFlatSpec with Matchers {
   }
 
   it should "prefer custom decoders over derived ones" in {
-    import auto._
```
satabin, line 43: These auto imports should no be removed, they are what is under test (the fact they their are not used)

csv/shared/src/main/scala/fs2/data/csv/CellDecoder.scala
```diff
@@ -191,10 +191,10 @@ object CellDecoder extends CellDecoderInstances1 with LiteralCellDecoders with E
       case d                  => new DecoderError(s"parsed duration isn't finite: $d").asLeft
     }
 
-  implicit val javaUrlDecoder: CellDecoder[URL] = s =>
+  implicit val javaUrlDecoder: CellDecoder[URI] = s =>
```
satabin, line 194: I wouldn't remove the `URL` one, since it's part of the API. Maybe just add the `URI` variation and make the `URL` one use it.

satabin (changes requested):
Some minor comments but otherwise it looks good. Thanks for this contribution!

csv/shared/src/main/scala/fs2/data/csv/CellDecoder.scala
cquiroz: Is it ok to do `CellDecoder/Encoder[URL]` only for the jvm side? otherwise the tests won't link on the scala.js side?

satabin: Sure, this is fine.

cquiroz:
I included the suggested changes

csv/js/src/main/scala/fs2/data/csv/PlatformCellEncoders.scala
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package fs2.data.csv
+
+trait PlatformCellEncoders {}
```
satabin, line 18: For the sake of symmetry, don't we want to have the `URI` encode as well?

cquiroz: The URI encoder is in the shared `CellEncoder` class but I could move it here if you prefer

satabin: Oh I didn't see it, right. This looks good, thanks.

satabin (approved):

satabin:
I just published a `0.8.0-SNAPSHOT` with this PR merged, if you want to give it a try.

cquiroz:
I can't find it any repo, should it be in sonatype snapshots?

satabin:
That's where I pushed it :thinking: I don't know if it needs some time to replicate for the snaphsots repo as well, but I also cannot find it on sonatype snapshots repo. I will wait a few hours to see if it appears, and if not I'll have another look tomorrow.

satabin:
Ok, I found it, I made a typo in `SNAPSHOT` so it was published to a staging repo, waiting to be released. Now it should be there for you to test.

cquiroz:
Thanks now I got it and it is working fine on my cross compiled project

satabin:
Happy to read that!

PR #66, opened 2020-10-14
Add JSON context in when using selectors

satabin:
This PR intentionally restricts the context to filtering and transformation. It induces overhead which might degrade general parsing too much and even prevent from parsing deeply nested JSON values. Restricting to selectors is ok as the context will be as deep as the selector, independently of the JSON value.

### DIFF
--- a/json/shared/src/main/scala/fs2/data/json/JsonContext.scala
+++ b/json/shared/src/main/scala/fs2/data/json/JsonContext.scala
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2020 Lucas Satabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package fs2
+package data
+package json
+
+import cats.Show
+import cats.implicits._
+
+/** A JSON trace context, positioning the current location in the JSON value
+  * starting from the root.
+  */
+sealed trait JsonContext
+
+object JsonContext {
+  case object Root extends JsonContext
+  case class Key(name: String, parent: JsonContext) extends JsonContext
+  case class Index(idx: Long, parent: JsonContext) extends JsonContext
+
+  implicit object JsonContextShow extends Show[JsonContext] {
+    def show(t: JsonContext): String = t match {
+      case Root => "/"
+      case Key(name, Root) => show"/$name"
+      case Key(name, parent) => show"$parent/$name"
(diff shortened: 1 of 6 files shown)


json/shared/src/main/scala/fs2/data/json/selectors.scala
```diff
@@ -93,6 +96,18 @@ object Selector {
...
   }
+
+  implicit object SelectorShow extends Show[Selector] {
+    def show(t: Selector): String = t match {
+      case ThisSelector => "."
+      case NameSelector(pred, strict, mandatory) =>
+        show"$pred${if (strict) "" else "?"}${if (mandatory) "!" else ""}"
+      case IndexSelector(pred, strict) => show"$pred${if (strict) "" else "?"}"
+      case IteratorSelector(strict)    => show".[]${if (strict) "" else "?"}"
+      case PipeSelector(left, right)   => show"$left$right"
```
ybasket, line 107: Isn't the pipe missing here?

json/shared/src/main/scala/fs2/data/json/selectors.scala
```diff
@@ -138,4 +161,12 @@ object IndexPredicate {
   case class Range(start: Int, end: Int) extends IndexPredicate {
     def apply(idx: Int): Boolean = start <= idx && idx <= end
   }
+
+  implicit val IndexPredicateShow: Show[IndexPredicate] = Show.show {
```
ybasket, line 165: Minor: For consistency, maybe use always `implicit object`?

json/shared/src/main/scala/fs2/data/json/JsonContext.scala
```diff
@@ -0,0 +1,42 @@
...
+  * starting from the root.
+  */
+sealed trait JsonContext
+
+object JsonContext {
+  case object Root extends JsonContext
+  case class Key(name: String, parent: JsonContext) extends JsonContext
+  case class Index(idx: Long, parent: JsonContext) extends JsonContext
+
+  implicit object JsonContextShow extends Show[JsonContext] {
```
ybasket, line 33: I find it cool to have `Show` instances for all these and I'm wondering whether it's worth to have instances of `Eq` too?

json/shared/src/main/scala/fs2/data/json/JsonContext.scala
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2019 Lucas Satabin
```
ybasket, line 2: Super-minor: 2020?
I'm still in favor of having a sbt plugin handle these headers btw

ybasket (changes requested):
Looks good, just a few minor comments and nitpicks

json/shared/src/main/scala/fs2/data/json/selectors.scala
satabin: IIRC I did it this way because of some error of the typer complaining the implicit lacked explicit return type when looked up in the `Selector` one.

json/shared/src/main/scala/fs2/data/json/JsonContext.scala
satabin: I want this to be used for reporting purpose, I am not convinced whether `Eq` is important in such use cases.

ybasket (approved):
Your comments sound all reasonable, so LGTM once the checks succeed!

PR #68, opened 2020-10-15
Add CBOR Support

satabin:


satabin:
@ybasket This is pretty much ready to merge. I created #102 for the CBOR <-> JSON integration. This implements the base features.

build.sbt
```diff
@@ -8,14 +8,15 @@ val commonSettings = List(
   scalaVersion := scala212,
   crossScalaVersions := Seq(scala213, scala212),
   organization := "org.gnieh",
+  headerLicense := Some(HeaderLicense.ALv2("2020", "Lucas Satabin")),
```
ybasket, line 11: 2021 by now? The files in this PR have a mix 🤨

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
```diff
@@ -0,0 +1,321 @@
...
+        case Some((hd, tl)) => parseIndefiniteArray(hd, 0, tl, acc, Nil)
+        case None           => Pull.raiseError(new CborException("unexpected end of input"))
+      }
+    } else {
+      chunk(idx) match {
+        case CborItem.Break =>
+          Pull.pure((chunk, idx + 1, rest, chunkAcc, CborValue.Array(acc.result(), true)))
+        case _ =>
+          if (acc.size == Int.MaxValue) {
+            raise(new CborException(s"array size is limited to max int (${Int.MaxValue}) elements"), chunkAcc)
```
ybasket, line 85: If this exception is ever thrown somewhere, we know fs2-data sees proper production usage!

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
```diff
@@ -0,0 +1,321 @@
...
+      }
+    }
+
+  private def parseTags[F[_]](chunk: Chunk[CborItem],
+                              idx: Int,
+                              rest: Stream[F, CborItem],
+                              tags: CborValue => CborValue,
+                              chunkAcc: List[CborValue])(implicit
+      F: RaiseThrowable[F]): Pull[F, CborValue, Result[F, CborValue => CborValue]] =
+    if (idx >= chunk.size) {
```
ybasket, line 213: I always wonder whether this could be extract to an `@inline` method to keep the code DRY.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
```diff
@@ -0,0 +1,321 @@
...
+          case None           => Pull.raiseError(new CborException("unexpected end of input"))
+        }
+      } else {
+        chunk(idx) match {
+          case CborItem.False     => Pull.pure((chunk, idx + 1, rest, chunkAcc, tags(CborValue.False)))
+          case CborItem.True      => Pull.pure((chunk, idx + 1, rest, chunkAcc, tags(CborValue.True)))
+          case CborItem.Null      => Pull.pure((chunk, idx + 1, rest, chunkAcc, tags(CborValue.Null)))
+          case CborItem.Undefined => Pull.pure((chunk, idx + 1, rest, chunkAcc, tags(CborValue.Undefined)))
+          case CborItem.SimpleValue(value) =>
+            Pull.pure((chunk, idx + 1, rest, chunkAcc, tags(CborValue.SimpleValue(value))))
```
ybasket, line 242: If you bind the pattern `CborItem.SimpleValue(value)` to a variable, you can re-use it instead of allocating a new one here (same in some other cases). Not big, but very easy to do and on the hot path.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueSerializer.scala
```diff
@@ -0,0 +1,107 @@
...
+          case Some((hd, tl)) => go(hd, 0, tl, Nil)
+          case None           => Pull.done
+        }
+      } else {
+        chunk(idx) match {
+          case CborValue.Integer(i) =>
+            // if representation is too big, use the tags specified in the RFC
+            if (i >= 0) {
+              val bytes = ByteVector(i.toByteArray)
+              if (bytes.size <= 64)
```
ybasket, line 49: 64 _bytes_? I guess you have the green RFC tests, but I would have expected bits.

cbor/shared/src/main/scala/fs2/data/cbor/low/internal/ItemParser.scala
```diff
@@ -0,0 +1,378 @@
...
+      case 21 =>
+        Pull.pure((chunk, idx, rest, CborItem.True :: chunkAcc))
+      case 22 =>
+        Pull.pure((chunk, idx, rest, CborItem.Null :: chunkAcc))
+      case 23 =>
+        Pull.pure((chunk, idx, rest, CborItem.Undefined :: chunkAcc))
+      case 24 =>
+        requireOneByte(chunk, idx, rest, chunkAcc).flatMap { case (chunk, idx, rest, chunkAcc, byte) =>
+          if (byte >= 0 && byte < 32)
+            Pull.raiseError(new CborException(s"invalid simple value additional byte $byte"))
```
ybasket, line 310: There's only `CborException` right now, right? Would it maybe make sense to introduce two types to differentiate between errors on the low-level parsing vs errors on the high-level parsing? Not that I can come up with a compelling use case right now, but it wouldn't harm and is hard/impossible to change once bincompat guarantees are given.

ybasket (approved):
Basically GTM, only some minor comments.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
satabin: Not really, it's not the same type, as this converts from items to values :)

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueSerializer.scala
satabin: I used to encode as `BitVector` and forgot to change this check. Thanks.

cbor/shared/src/main/scala/fs2/data/cbor/low/internal/ItemParser.scala
satabin: I am generally speaking not a big fan of exception proliferation. I would see a reason to split between parsing and deserialization errors, but I don't want to create too many of them for parsing.

build.sbt
satabin: Well, this branch started in 2020, and I didn't upgrade ^^

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
satabin: I thought about it and didn't find anything satisfying so far.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
ybasket: Details^^
Yeah, I see it now as well!

cbor/shared/src/main/scala/fs2/data/cbor/low/internal/ItemParser.scala
ybasket: I don't mean one for every possible error, but at least introducing a `sealed trait CborException` already with two subtypes. If we then discover valid use cases for more detailed errors, we can add them within bincompat.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueSerializer.scala
ybasket: Make sure to check the whole method, there are some more magic numbers in this method that look suspiciously like bit counts.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
ybasket: Fancy way: Use `ContT`. But it feels like it could also be done within the monadic context of `Pull`, so things become more `flatMap`py and can maybe even be written as for-comprehension.

cbor/shared/src/main/scala/fs2/data/cbor/low/internal/ItemParser.scala
satabin: Meh, why not.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
satabin: Trying out something like this:

```scala
  def ensureChunk[F[_], T](chunk: Chunk[Byte], idx: Int, rest: Stream[F, Byte], chunkAcc: List[CborItem])(
      cont: (Chunk[Byte], Int, Stream[F, Byte], List[CborItem]) => Pull[F, CborItem, T])(
      onEos: => Pull[F, CborItem, T]): Pull[F, CborItem, T] =
    if (idx >= chunk.size) {
      Pull.output(Chunk.seq(chunkAcc)) >> rest.pull.uncons.flatMap {
        case Some((hd, tl)) => cont(hd, 0, tl, Nil)
        case None           => onEos
      }
    } else {
      cont(chunk, idx, rest, chunkAcc)
    }
```

I will need to tune the compiler settings to inline it, though

satabin: I pushed something, if you could have a look at it it would be great.

PR #103, opened 2021-02-05
Fix recursive chunk `uncons`

satabin:
When a chunk is pulled, we need to recursively ensure that it is not
empty in a stack safe way.

### DIFF
--- a/cbor/shared/src/main/scala/fs2/data/cbor/low/internal/ItemParser.scala
+++ b/cbor/shared/src/main/scala/fs2/data/cbor/low/internal/ItemParser.scala
@@ -42,7 +42,7 @@ private[low] object ItemParser {
       onEos: => Pull[F, CborItem, T]): Pull[F, CborItem, T] =
     if (idx >= chunk.size) {
       Pull.output(Chunk.seq(chunkAcc.reverse)) >> rest.pull.uncons.flatMap {
-        case Some((hd, tl)) => cont(hd, 0, tl, Nil)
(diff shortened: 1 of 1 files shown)


cbor/shared/src/main/scala/fs2/data/cbor/low/internal/ItemParser.scala
```diff
@@ -42,7 +42,7 @@ private[low] object ItemParser {
       onEos: => Pull[F, CborItem, T]): Pull[F, CborItem, T] =
     if (idx >= chunk.size) {
       Pull.output(Chunk.seq(chunkAcc.reverse)) >> rest.pull.uncons.flatMap {
-        case Some((hd, tl)) => cont(hd, 0, tl, Nil)
+        case Some((hd, tl)) => ensureChunk(hd, 0, tl, Nil)(cont)(onEos)
```
ybasket, line 45: One thing I started wondering: Is it safe to assume `idx=0` here? Or could it be that we need to do something like `idx - chunk.size` here to be safe (for future developments)?

ybasket (approved):

cbor/shared/src/main/scala/fs2/data/cbor/low/internal/ItemParser.scala
satabin: This is an interesting topic, that would allow to have built-in selector in the parser. However, this will required to rethink the entire parser, so I prefer to not do it now, and think about it as a new implementation in the future.

Issue #44, opened 2020-07-16
Rethink and rework the CSV pipes


satabin:
The CSV module was the first one written in this project and suffers from a really cluttered and surprising API. For the 1.0 release, let's rework it completely to make it easier to use.

PR #104, opened 2021-02-06
Introduce RowF abstraction

ybasket:
To keep code in the CSV module DRY, this PR introduces `RowF`, a class that generalizes over the presence of headers in CSV rows via generic parameters. Both `Row` and `CsvRow` are now just type aliases of `RowF`, parameterized according to their needs. This avoid the repetition and inheritance that was caused by these two.

While not binary compatible, this change should be source-compatible in (almost) all practical use cases, at worst a `new Row(...)` becomes `Row(...)`.

Documentation is adjusted to compile and be correct, an in-depth explanation is due together with the new pipes from #44 that build on top of the changes here.


satabin (changes requested):
GTM, you should probably also update the plugins so that mdoc can wok with scala 2.13.4

ybasket:
> GTM, you should probably also update the plugins so that mdoc can wok with scala 2.13.4

Ah, I assumed it was on the latest already due to Scala Steward and hoped it would run fine on Travis. But yeah, 1.x branch wasn't fully up-to-date, now it works fine. Thanks!

satabin (approved):

PR #126, opened 2021-03-02
#45: Add @CsvEmbed

### LINKED ISSUE #45: Support nested case classes for CsvRow*coder derivation
When deriving shapeless implicits for very wide case classes, sometimes the compiler fails with a stack overflow error. One technique to avoid this is to nest the case classes, which is isomorphic to the flattened structure.

Something like this: 
```scala
case class Sub(
  @CsvName("first-name") firstName: String,
  @CsvName("the-age") age: Int
)
case class Whole(
  @CsvName("last-name") lastName: String,
  @CsvEmbed sub: Sub
)
// last-name, first-name, the-age
// flintstone, fred, 42
```

Context: https://gitter.im/fs2-data/general?at=5f159b91a28d973192e7d07a

Doobie supports a similar structure

ybasket:
Allows to map nested case class structures from and to flat CSV. See included docs for more info. Fixes #45.

### DIFF
--- a/csv/generic/src/main/scala/fs2/data/csv/generic/MapShapedCsvRowDecoder.scala
+++ b/csv/generic/src/main/scala/fs2/data/csv/generic/MapShapedCsvRowDecoder.scala
@@ -27,51 +27,79 @@ trait MapShapedCsvRowDecoder[Repr] extends CsvRowDecoder[Repr, String]
 
 object MapShapedCsvRowDecoder extends LowPriorityMapShapedCsvRowDecoder1 {
 
-  implicit def hnilRowDecoder[Wrapped]: WithDefaults[Wrapped, HNil, HNil, HNil] =
-    new WithDefaults[Wrapped, HNil, HNil, HNil] {
-      def fromWithDefault(row: CsvRow[String], default: HNil, annotation: HNil): DecoderResult[HNil] =
-        Right(HNil)
-    }
+  implicit def hnilRowDecoder[Wrapped]: WithDefaults[Wrapped, HNil, HNil, HNil, HNil] =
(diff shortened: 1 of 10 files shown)


csv/generic/src/main/scala/fs2/data/csv/generic/MapShapedCsvRowDecoder.scala
```diff
@@ -27,80 +27,142 @@ trait MapShapedCsvRowDecoder[Repr] extends CsvRowDecoder[Repr, String]
...
+                     None.type :: NamesTail,
+                     Some[CsvEmbed] :: EmbedTail] =
+    (row: CsvRow[String],
+     default: Option[Option[Head]] :: DefaultTail,
+     names: None.type :: NamesTail,
+     embeds: Some[CsvEmbed] :: EmbedTail) => {
+      for {
+        head <- (Head(row), default.head) match {
+          case (r @ Right(_), _)                                    => r
+          case (Left(_: DecoderError.ColumnMissing), Some(default)) => Right(default)
```
ybasket, line 86: This is a somewhat tricky case as it could be that some of the fields of the embedded value are present, some are invalid. In this edge case, it depends on the order of fields whether the default is taken or the error of an invalid field is bubbled up. The only way to mitigate this would be to introduce error-accumulating parsing (which would indeed be an interesting addition), but as the exact combination of 
* embedding 
* invalid fields for the embedded value 
* missing fields for the embedded value and
* a default for the embedded field 

should be extremely rare in the wild, I believe this is fine.

satabin (changes requested):
Looks good, but you probably need to rebase to get the line context into account.

PR #138, opened 2021-03-29
Check formatting for PRs

satabin:


cbor/shared/src/main/scala/fs2/data/cbor/Tags.scala
```diff
@@ -22,24 +22,24 @@ package cbor
   */
 object Tags {
 
-  final val StandardDateTimeString: Long = 0L
-  final val EpochDateTimeString: Long = 1L
-  final val PositiveBigNum: Long = 2L
-  final val NegativeBigNum: Long = 3L
-  final val DecimalFraction: Long = 4L
-  final val BigFloat: Long = 5L
+  final val StandardDateTimeString: Long = 0l
```
ybasket, line 25: Can we set `literals.long=Upper` for scalafmt? I find that clearly more readable and according to the docs, it is even the default?

ybasket (approved):
One comment, otherwise LGTM

cbor/shared/src/main/scala/fs2/data/cbor/Tags.scala
satabin: I thought I did. I will make it happen now.

PR #145, opened 2021-04-03
New CSV pipes

### LINKED ISSUE #44: Rethink and rework the CSV pipes

ybasket:
New pipes are there as planned, lacks docs and tests. More description to follow. 

Closes #44

### DIFF
--- a/csv/shared/src/main/scala/fs2/data/csv/package.scala
+++ b/csv/shared/src/main/scala/fs2/data/csv/package.scala
@@ -100,98 +100,206 @@ package object csv {
     case object Literal extends QuoteHandling
   }
 
-  /** Transforms a stream of characters into a stream of CSV rows.
-    *
-    * @param separator character to use to separate fields in the CSV
-    * @param quoteHandling use [[QuoteHandling.RFCCompliant]] for RFC-4180
-    *                      handling of quotation marks (optionally quoted
-    *                      if the value begins with a quotation mark; the
-    *                      default) or [[QuoteHandling.Literal]] if quotation
-    *                      marks should be treated literally
-    */
-  def rows[F[_], T](separator: Char = ',', quoteHandling: QuoteHandling = QuoteHandling.RFCCompliant)(implicit
-      F: RaiseThrowable[F],
-      T: CharLikeChunks[F, T]): Pipe[F, T, NonEmptyList[String]] =
-    RowParser.pipe[F, T](separator, quoteHandling)
+  def decodeWithoutHeaders[F[_], C, T]: PartiallyAppliedDecodeWithoutHeaders[T] =
+    new PartiallyAppliedDecodeWithoutHeaders[T](dummy = true)
 
-  /** Transforms a stream of raw CSV rows into parsed CSV rows with headers. */
-  def headers[F[_], Header](implicit
-      F: RaiseThrowable[F],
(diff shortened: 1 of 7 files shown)


ybasket:
@satabin It lacks doc updates (only Scaladoc so far), but before I do that, maybe you could have a look to see whether you agree with the API?

csv/shared/src/main/scala/fs2/data/csv/package.scala
```diff
@@ -100,98 +100,238 @@ package object csv {
...
-    * @param separator character to use to separate fields in the CSV
-    * @param quoteHandling use [[QuoteHandling.RFCCompliant]] for RFC-4180
-    *                      handling of quotation marks (optionally quoted
-    *                      if the value begins with a quotation mark; the
-    *                      default) or [[QuoteHandling.Literal]] if quotation
-    *                      marks should be treated literally
+  /** Decode a char-like stream (see [[CharLikeChunks]]) into a specified type,
+    * assuming the file neither contains headers nor are they needed for decoding.
+    */
+  def decodeWithoutHeaders[F[_], C, T]: PartiallyAppliedDecodeWithoutHeaders[T] =
```
satabin, line 106: `F` and `C` are too much, I guess?

satabin (changes requested):
API looks good to me. I like that you kept the low level one as opt-in.

csv/shared/src/main/scala/fs2/data/csv/package.scala
ybasket: That's true, good catch!

documentation/docs/csv/index.md
```diff
@@ -227,14 +227,47 @@ implicit object MyRowDecoder extends CsvRowDecoder[MyRow, String] {
...
-val decoded = withh.through(decodeRow[Fallible, String, MyRow])
+val decoded = withh.through(lowlevel.decodeRow[Fallible, String, MyRow])
 decoded.compile.toList
 ```
 
 Analogously, you can encode your data with a `CsvRowEncoder`. Make sure to not vary the headers based on the data itself as this can't be reflected in the CSV format.
 
 As you can see this can be quite tedious to implement. Lucky us, the `fs2-data-csv-generic` module comes to the rescue to avoid having to write the boilerplate. Please refer to [the module documentation][csv-generic-doc] for more details.
 
+### High-level API for standard use cases
```
satabin, line 238: I would move this section up, before the low-level one, as it covers the standard use cases, let's see it first.

I think it's better for users to have it first, and then low-level saying: if the previous thing doesn't cover your needs, then have a look at low-level, ...

satabin (changes requested):
Code looks good, just a documentation thing I'd like to change. Otherwise LGTM

documentation/docs/csv/index.md
ybasket: I added an explanation of the API split and a quick-start using generic as that is what most people will be using. It has to forward to later sections for the type classes, but I believe it gives a good first impression of what the library offers and is easy to understand for people familiar with circe or similar. Does that fit your expectations?

satabin (approved):
Thanks for taking my comments into account. This looks super good.

PR #147, opened 2021-04-05
Rename overloads in RowF

ybasket:
Prepares for changes in overload resolution in Scala 3(-RC2) which make these overloaded methods very annoying to implement and hard to call. Even though we have no Scala 3 support yet, it avoids a breaking change or Scala version-dependent API difference in fs2-data's 1.0.


satabin (approved):
LGTM

PR #152, opened 2021-04-11
Add some convenience CSV coders

ybasket:
Instances that come for free as their underlying collections are isomorphic to rows. Useful in combination with the new CSV pipes.


satabin (approved):
LGTM

PR #150, opened 2021-04-08
Scala 3

ybasket:
- [x] Add Scala `3.0.0-RC2` to build
- [x] ~Switch `contextual` for `literally`~ done in #156, needs rebase 
- [x] Have a `diffson` version on Scala 3. PR is open: https://github.com/gnieh/diffson/pull/227
- [x] Fix problems coming from lack of `bm4`
- [x] Add a Scala 3 variant of `csv-generic`
- [x] Document Scala 3 derivation limitations
- [x] Get mdoc to compile again
- [x] Re-enable scalafmt on CI

PR #156, opened 2021-04-16
Replace contextual with literally

ybasket:
As contextual isn't moving towards Scala 3 (yet), this PR replaces it with the smaller, more focused and perfectly cross-building typelevel/literally lib.

A modified extract from #150 to make this change before the next RC.


satabin (approved):
This looks good, thanks a lot for this.

PR #182, opened 2021-05-15
Drop need for better-monadic-for

ybasket:
Rewrites all for-comprehensions that require bm4 to to compile into explicit flatMap chains. That helps transitioning to Scala 3 where there is (so far) no support for extracting tuples in for-comprehensions. Not removing bm4 from the build as Scala 2 will still profit from tail-map elimination in the remaining for-comprehensions.

Rewrite was done using IDEA, probably one could manually make some of the `flatMap` stacks nicer (I already optimised formatting though).

### DIFF
--- a/xml/shared/src/main/scala/fs2/data/xml/internals/EventParser.scala
+++ b/xml/shared/src/main/scala/fs2/data/xml/internals/EventParser.scala
@@ -260,11 +260,12 @@ private[xml] object EventParser {
           } else {
             T.current(ctx) match {
               case '/' =>
-                for {
-                  (ctx, chunkAcc, qname) <- readQName(T.advance(ctx), chunkAcc)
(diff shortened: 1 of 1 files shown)


xml/shared/src/main/scala/fs2/data/xml/internals/EventParser.scala
```diff
@@ -624,8 +625,7 @@ private[xml] object EventParser {
                   sb.append('>')
                 } else {
                   sb.append('&')
-                  for (i <- 0 until n)
-                    sb.append("gt;".charAt(i))
+                  sb.append("gt;".substring(0, n))
```
ybasket, line 628: I guess that's the only "real" change in this PR.

satabin: Wow, why didn't I do it this way already? :facepalm: Thanks

PR #206, opened 2021-07-08
Cache the `byHeader` field

### LINKED ISSUE #205: CellDecoder performance
I tried to use the `data-csv` library to parse some large files. Reading them into the `RowF` case class was quite fast but I did get some poor performance during the cell decoder stage. It did occur for both the semi-auto and custom-written decoders. I've located a code path in the `RowF.as[T]` function where it does recreate the `byHeader` value for each field that's extracted. I could not think of a fix for it without any breaking API changes. I'm happy to provide a pull request if we can find a suitable solution.

A temporary workaround for me was to create a wrapper and cache the `byHeader` value and provide a custom decoder. Here's an example:
```
class CachedHeaderRow(val origin: csv.RowF[Some, String])(implicit hasHeaders: HasHeaders[Some, String]) {

  private lazy val byHeader = origin.headers.get.toList.zip(origin.values.toList).toMap

  def as[T](header: String)(implicit decoder: CellDecoder[T]): DecoderResult[T] =
    byHeader.get(header) match {
      case Some(v) => decoder(v)
      case None    => Left(new DecoderError(s"unknown field $header"))
    }
}
```

This decreased the pipeline of reading, parsing, and decoding of 1 000 000 rows with 30 cells from ~100 seconds to ~10 seconds.

satabin:
Having it as a def allowed for statically checking that headers are
defined and avoid accidental call to it when no headers are defined.
However, this implies that the map is recomputed for each access to a
function like `as`, which induces significant overhead.

Turning it into a lazy has these implications:
 - significantly improve performance when transforming rows with lots of
   cells
 - calling this field becomes more unsafe since there is no static check
   anymore that the headers are actually defined

However the risk is mitigated by the fact that this private field is
called by a quite stable API which has the constraint, ensuring the call
is safe and by adding a deprecation warning so that new uses will at
least be made aware of having to check things.

Fixes #205

### DIFF
--- a/csv/shared/src/main/scala/fs2/data/csv/RowF.scala
+++ b/csv/shared/src/main/scala/fs2/data/csv/RowF.scala
@@ -18,6 +18,7 @@ package fs2.data.csv
 import cats._
 import cats.data._
 import cats.implicits._
+import scala.annotation.nowarn
 
 /** A CSV row with or without headers. The presence of headers is encoded via the first type param
   * which is a subtype of [[scala.Option]]. By preserving this information in types, it's possible to define
@@ -112,14 +113,14 @@ case class RowF[H[+a] <: Option[a], Header](values: NonEmptyList[String],
     * An empty cell value results in `Some("")`.
     */
   def apply(header: Header)(implicit hasHeaders: HasHeaders[H, Header]): Option[String] =
-    byHeader.get(header)
+    byHeader.get(header): @nowarn("msg=HasHeaders")
 
   /** Returns the decoded content of the cell at `header`.
     * Fails if the field doesn't exist or cannot be decoded
     * to the expected type.
     */
   def as[T](header: Header)(implicit hasHeaders: HasHeaders[H, Header], decoder: CellDecoder[T]): DecoderResult[T] =
-    byHeader.get(header) match {
+    (byHeader: @nowarn("msg=HasHeaders")).get(header) match {
       case Some(v) => decoder(v)
       case None    => Left(new DecoderError(s"unknown field $header"))
     }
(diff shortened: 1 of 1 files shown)


csv/shared/src/main/scala/fs2/data/csv/RowF.scala
```diff
@@ -139,7 +140,13 @@ case class RowF[H[+a] <: Option[a], Header](values: NonEmptyList[String],
...
   def dropHeaders: Row = Row(values)
 
-  private def byHeader(implicit hasHeaders: HasHeaders[H, Header]): Map[Header, String] =
+  // let's cache this to avoid recomputing it for every call to `as` or similar method
+  // the `Option.get` call is safe since this field is only called in a context where a `HasHeaders`
+  // instance is provided, meaning that the `Option` is `Some`
+  // of course using a lazy val prevents us to make this constraint statistically checked, but
+  // the gain is significant enough to allow for this local unsafety
+  @deprecated("Have you checked that you have a `HasHeaders` instance in scope?")
+  private lazy val byHeader: Map[Header, String] =
```
ybasket, line 149: The change is definitely fine (thank you), but maybe we can protect it by indirection via a def? So we would keep `@inline def byHeader` to check the implicit evidence, but it's implementation would be backed by a `private lazy val dontUseMeByHeader`? In the end, it doesn't change anything if we always keep the constraint in mind when changing the class' internals, but I prefer safe over sorry…

ybasket (approved):
One suggestion, but already GTM

csv/shared/src/main/scala/fs2/data/csv/RowF.scala
satabin: This is my first implementation. And then the compilation overhead seemed not worth it to me. Even thought about having a private object holding this private field and having one public method. In the end, it's all the same, and I think the warning makes it _safe enough_ ™️

ybasket: OK, fair enough. LGTM then!

PR #253, opened 2021-11-01
Update fs2 + Scala 3.

rpiaggio:
Could I request a release with these updates? We need to get rid of the transitive dependency to `scalajs-dom` 1.0 that was in `fs2` < `3.2.0`.

Thank you!


ybasket (approved):
Thank you very much! I'll merge some of the other dependency update PRs and then a release should be on its way soon.

ybasket:
And if you need something even earlier, version `1.1.0+13-f84c9dc8-SNAPSHOT` from Sonatype snapshots contains your PR.

PR #282, opened 2021-12-31
Add various streaming transducers

satabin:
* [x] Copyless Streaming Tree Transducers which can be used to implement query languages over tree like data
* [x] Streaming Finite State Transducers which can be used for matching inputs and emitting some output
* [x] ~Streaming String Transducers which can be used for matching inputs and emitting some output as soon as possible~

### DIFF
--- a/transducers/shared/src/main/scala/fs2/data/stt/Assignment.scala
+++ b/transducers/shared/src/main/scala/fs2/data/stt/Assignment.scala
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2021 Lucas Satabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fs2.data.stt
+
+import cats.Show
+import cats.syntax.show._
+
+sealed trait Assignment[+V <: Variable, +C]
+sealed trait Reset[+C] extends Assignment[Nothing, C]
+object Reset {
+  def unapply[C](r: Reset[C]): Some[Reset[C]] =
+    Some(r)
+}
+object Assignment {
+  case class Empty(x: Variable.Normal) extends Reset[Nothing]
+  case class Hole(x: Variable.Normal) extends Reset[Nothing]
+  case class Char[C](x: Variable.Normal, c: C) extends Reset[C]
+  case class Subtree[C](x: Variable.Normal, open: C, close: C) extends Reset[C]
+  case class Append[V <: Variable](x: Variable.Normal, y: V) extends Assignment[V, Nothing]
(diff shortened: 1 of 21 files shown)


transducers/shared/src/main/scala/fs2/data/stt/Assignment.scala
```diff
@@ -0,0 +1,56 @@
...
+
+package fs2.data.stt
+
+import cats.Show
+import cats.syntax.show._
+
+sealed trait Assignment[+V <: Variable, +C]
+sealed trait Reset[+C] extends Assignment[Nothing, C]
+object Reset {
+  def unapply[C](r: Reset[C]): Some[Reset[C]] =
```
ybasket, line 25: Is this really needed? Shouldn't it be a irrefutable pattern anyways and get handled by the compiler?

transducers/shared/src/main/scala/fs2/data/stt/STT.scala
```diff
@@ -0,0 +1,263 @@
...
+          env.lookup[F, Expr0[Out]](x).flatMap(loop(_, acc))
+        case Expr0.Char(c) =>
+          F.pure(acc.append(c))
+        case Expr0.Subtree(open, sub, close) =>
+          loop(sub, acc.append(open)).map(_.append(close))
+        case Expr0.Concat(left, right) =>
+          for {
+            acc <- loop(left, acc)
+            acc <- loop(right, acc)
+          } yield acc
```
ybasket, line 151:
```suggestion
          loop(left, acc).flatMap(loop(right, _))
```

I find the for-comprehension with re-use of `acc` unnecessarily hard to read and it has the tail `map` problem on Scala 3 IIRC.

transducers/shared/src/main/scala/fs2/data/stt/Table.scala
```diff
@@ -0,0 +1,38 @@
...
+}
+
+object Table {
+
+  implicit object PartialFunctionTable extends Table[PartialFunction] {
+    def get[From, To](m: PartialFunction[From, To])(from: From): Option[To] =
+      m.lift(from)
+  }
+
+  implicit object MapTable extends Table[Map] {
```
ybasket, line 35: `Map <: MapOps <: PartialFunction`, so we might run into ambiguous implicit here. Maybe move the partial function instance into a low priority trait to be safe?

ybasket (commented):
Not entirely sure what it does without docs, but left some small Scala-specific comments and will stay tuned for updates 😄

transducers/shared/src/main/scala/fs2/data/stt/Table.scala
satabin: > Not entirely sure what it does without docs, but left some small Scala-specific comments and will stay tuned for updates smile

Documentation will come. This is a first basic building block for higher-level features. More transducers to come soon, I hope, especially SSTs, which will allow for parsing. This STT can be the base for streaming query languages on tree structures.

transducers/shared/src/main/scala/fs2/data/stt/Assignment.scala
satabin: How do you plan on writing `case Reset(r) =>` if there is no `unapply` method in object `Reset`? This is basically what the compiler would generate if `Reset` were a case class, IIRC.

ybasket: Ah, ignore me. I confused myself while jumping between the trait and the `case class`es implementing it. Hope it works fine on Scala 3 and it changes to `unapply` / introduction of `Matchable`.

PR #286, opened 2022-01-12
Cross-build JSON interpolators module to scala.js

ybasket:
Noticed we don't cross-build it though we can thanks to https://index.scala-lang.org/artifacts/satabin/fs2-data


satabin (approved):
Must be some legacy from when we were using contextual.

PR #309, opened 2022-03-30
Run more tests on both JVM/JS

armanbilge:
fs2-io supports all APIs (including files) on JVM and JS now.


armanbilge:
All the sources are shared now, so you can simplify your directory structure to an ordinary project (no need for shared/jvm/js split).

build.sbt
```diff
@@ -280,14 +280,17 @@ lazy val jsonInterpolators = crossProject(JVMPlatform, JSPlatform)
   .dependsOn(json % "compile->compile;test->test")
 
 lazy val xml = crossProject(JVMPlatform, JSPlatform)
-  .crossType(CrossType.Full)
+  .crossType(CrossType.Pure)
```
armanbilge, line 283: This setting controls the directory structure.

armanbilge:
@satabin should I fix these modules as well?
https://github.com/satabin/fs2-data/blob/bf5a7033574d17c05cf408ef33a93b7792ec1d8a/build.sbt#L250-L251
https://github.com/satabin/fs2-data/blob/bf5a7033574d17c05cf408ef33a93b7792ec1d8a/build.sbt#L264-L265

satabin:
Ok thanks I need to do a global check on my use of `shared`,  I thought I had more specific cases.

ybasket:
I let two of you figure out the sbt sources setup, it's by far not my strength 🙄 

But thank you very much for the work @armanbilge! Feel free to ping me on the Typelevel Discord (same handle) if you need more workflow approvals on the way.

build.sbt
```diff
@@ -57,7 +57,8 @@ val commonSettings = List(
     "co.fs2" %%% "fs2-core" % fs2Version,
     "org.scala-lang.modules" %%% "scala-collection-compat" % "2.7.0",
     "io.circe" %%% "circe-parser" % circeVersion % "test",
-    "co.fs2" %% "fs2-io" % fs2Version % "test",
+    "io.circe" %%% "circe-jawn" % "0.15.0-M1" % "test",
```
ybasket, line 60:
```suggestion
    "io.circe" %%% "circe-jawn" % circeVersion % "test",
```

Or is there any reason we need the milestone instead of the stable `0.14.1` that's used elsewhere?

armanbilge: Yes, to get the circe-jawn parser on JS. It will be available in 0.14.2 when that's released.

satabin: I’m fine with that since it’s for testing purpose.

ybasket (approved):
Looks good from my side then, but I'll let @satabin do the merge in case there's something with the source folders still to be optimised.

satabin:
> Looks good from my side then, but I'll let @satabin do the merge in case there's something with the source folders still to be optimised.

Nah, it’s fine, I’ll merge it. Thanks a lot @armanbilge!

PR #325, opened 2022-06-03
Make it possible to keep comments

satabin:
Comments can be useful to keep when transforming XML from source to
source.

### DIFF
--- a/xml/src/main/scala/fs2/data/xml/package.scala
+++ b/xml/src/main/scala/fs2/data/xml/package.scala
@@ -29,11 +29,24 @@ package object xml {
 
   /** Transforms a stream of characters into a stream of XML events.
     * Emitted tokens are guaranteed to be valid up to that point.
-    * If the streams ends without failure, the sequence of tokens is sensured
+    * If the streams ends without failure, the sequence of tokens is ensured
     * to represent a (potentially empty) sequence of valid XML documents.
     */
+  @deprecated(message = "Use `fs2.data.xml.events()` instead.", since = "fs2-data 1.4.0")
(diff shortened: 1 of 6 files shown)


xml/src/main/scala/fs2/data/xml/package.scala
```diff
@@ -29,11 +29,24 @@ package object xml {
...
+  /** Transforms a stream of characters into a stream of XML events.
+    * Emitted tokens are guaranteed to be valid up to that point.
+    * If the streams ends without failure, the sequence of tokens is ensured
+    * to represent a (potentially empty) sequence of valid XML documents.
+    *
+    * if `skipComment` is `false`, then comment event will be emitted
+    * together with the comment content.
+    */
+  def events[F[_], T](
+      skipComment: Boolean = true)(implicit F: RaiseThrowable[F], T: CharLikeChunks[F, T]): Pipe[F, T, XmlEvent] =
```
ybasket, line 48: At least my mental model prefers something like `includeComments` with a default of `false`, wdyt?

ybasket (approved):
One comment on naming, apart from that it looks good!

xml/src/main/scala/fs2/data/xml/package.scala
satabin: I was thinking in "true by default", but sure.

ybasket: I come from "opt-in to extra work" and if true, then comment.isDefined (pseudo) also being true.

Both views make sense to me, so merge as you think is best.

satabin: I made it use less negation, it probably improves code readablity, thanks.

PR #335, opened 2022-06-14
Fix `accept` when crossing chunk boundary

satabin:
Looping on `accept` instead of `loop` restarts reading the string to
accept from the beginning, which results in wrongly reported syntax
error.

Thanks @armanbilge for spotting this bug.


ybasket (approved):

armanbilge:
I believe this was @rossabaker's 🔎 🐛  but 🎉  all the same! :)

satabin:
> I believe this was @rossabaker's mag_right bug but tada all the same! :)

You’re right, sorry for the mis-attribution.

satabin:
> > It looks like since upgrading fs2, Scala.JS tests are a bit flaky.
> 
> Wooops! Can you point me to something specific? I did a big refactor of the internal implementation and probably introduced some bugs sweat

I was about to ping you :sweat_smile: 
The CI regularly fails like [this one](https://github.com/satabin/fs2-data/actions/runs/2494766954). Locally it looks stable, this is weird. I am testing whether this is the update that introduced the problem.

armanbilge:
Oh dear, thanks, that looks pretty ominous 😱 let's see if I can at least replicate it locally.

satabin:
@armanbilge it looks like downgrading did the trick. I will release a bugfix release with it. If you need more inputs on my problems, let me know.

PR #336, opened 2022-06-15
Add CBOR/JSON interoperability

satabin:
- [X] Add CBOR items to JSON tokens transformation
 - [x] Test correct behavior regarding decoding tags
   - [x] Make sure strings are properly decoded
   - [x] Make sure tag is propagated to array values
   - [x] Make sure tag is propagated to map key and values
 - [X] Add diagnostic pipe
 - [x] Add JSON tokens to CBOR items transformation
 - [x] Add documentation

### DIFF
--- a/cbor-json/shared/src/test/scala/fs2/data/cbor/CborToJsonSpec.scala
+++ b/cbor-json/shared/src/test/scala/fs2/data/cbor/CborToJsonSpec.scala
@@ -4,7 +4,7 @@ package cbor
 
 import low.{items, CborItem}
 import fs2.data.json.circe._
-import fs2.data.json.Token
+import fs2.data.json.{Token, ast}
 import fs2.io.file.{Files, Path}
 
 import weaver._
@@ -35,11 +35,11 @@ object CborToJsonSpec extends SimpleIOSuite {
             Stream
               .chunk(Chunk.byteVector(cbor))
               .through(items[IO])
-              .through(fs2.data.cbor.json.decodeItems)
-              .through(fs2.data.json.ast.values)
+              .through(json.decodeItems)
+              .through(ast.values)
               .compile
               .lastOrError
-              .map(s => expect(s == expected).or(expect(Some(s) == expectedAlt, s"failed test at index $idx")))
+              .map(s => expect.same(expected, s).or(expect.same(expectedAlt, Some(s))))
           }
           .compile
           .foldMonoid)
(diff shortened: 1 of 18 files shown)


cbor/shared/src/main/scala/fs2/data/cbor/Tags.scala
```diff
@@ -40,6 +40,6 @@ object Tags {
   final val RegularExpression: Long = 35L
   final val MimeMessage: Long = 36L
 
-  final val SelfDescribeCbor: Long = 55799L
+  final val SelfDescribedCbor: Long = 55799L
```
ybasket, line 43: This is bincompat-breaking if I'm not mistaken. We can fix that by making the old one package private while adding the new one. Still source breaking, but that's less of a concern.

cbor/shared/src/main/scala/fs2/data/cbor/package.scala
```diff
@@ -0,0 +1,22 @@
+package fs2
+package data
+
+package object cbor {
+
+  /** Useful for debugging, generates a stream of diagnostic strings
+  * representing the CBOR values in the input stream as defined in
+  * [section 8 of RFC8949](https://www.rfc-editor.org/rfc/rfc8949.html#name-diagnostic-notation).
+  */
+  def diagnostic[F[_]](s: Stream[F, low.CborItem])(implicit F: RaiseThrowable[F]): Stream[F, String] =
```
ybasket, line 10: Any reason why this is no `Pipe`? It's effectively equivalent.

cbor/shared/src/test/scala/fs2/data/cbor/ParserSpec.scala
```diff
@@ -394,10 +398,14 @@ object ParserSpec extends SimpleIOSuite {
...
-          (low, high) =>
-            expect(low == Right(expectedLow), s"CBOR item parser should parse ${input.toHex} properly") and
-              expect(high == Right(expectedHigh), s"CBOR value parser should parse ${input.toHex} properly")
+        (bytes
+           .through(items[IO])
+           .compile
+           .toList
+           .attempt,
+         bytes.through(values[IO]).compile.toList.attempt).mapN { (low, high) =>
+          expect(low == Right(expectedLow), s"CBOR item parser should parse ${input.toHex} properly") and
```
ybasket, line 407: For all tests: Shall we switch to `expect.eql/same`? It conveys the intent better and has special rendering in case of mismatches…

documentation/docs/cbor-json/index.md
```diff
@@ -0,0 +1,94 @@
...
+    CborItem.TextString("Some string"),
+    CborItem.TextString("Some other string"),
+    CborItem.False
+  ))
+  .through(cbor.json.decodeItems[Fallible])
+  .compile
+  .toList
+```
+
+CBOR is more expressive than JSON and allows to express more kind of data, so no all CBOR item streams can be successfully converted, or some information loss might occur.
```
ybasket, line 41:
```suggestion
CBOR is more expressive than JSON and allows to express more kind of data, so not all CBOR item streams can be successfully converted, or some information loss might occur.
```

cbor-json/shared/src/test/scala/fs2/data/cbor/CborToJsonSpec.scala
```diff
@@ -0,0 +1,161 @@
...
+        (idx, cbor, json, jsonAlt)
+      })
+      .flatMap(
+        Stream
+          .emits(_)
+          .evalMap { case (idx, cbor, expected, expectedAlt) =>
+            Stream
+              .chunk(Chunk.byteVector(cbor))
+              .through(items[IO])
+              .through(fs2.data.cbor.json.decodeItems)
```
ybasket, line 38: Given the split package definition at the start of this file, this could be simplified to `json.decodeItems`. Or is the FQN intentional?

ybasket (approved):
Some minor comments, but looks good overall 🥳

PR #343, opened 2022-07-02
Add streaming DFA with basic XPath and JsonPath implementations

satabin:
This PR aims to generalize over querying semi structured data, with a common streaming DFA engine. It will only support basic queries (e.g. with **no** support for sub-queries) but with recursive results (e.g. `//` in XPath and `..` in JsonPath)

 - [x] Implement streaming DFA engine
   - [x] Make it symbolic
   - [x] Support predicate 
 - [x] Implement XPath
   - [x] support attribute predicates 
   - [x] write parser
 - [x] Implement JsonPath
   - [x] write parser
 - [x] Add documentation


finite-state/shared/src/main/scala/fs2/data/pfsa/PDFA.scala
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package fs2.data.pfsa
+
+import Pred.syntax._
+
+import cats.Show
+import cats.syntax.all._
+
+class PDFA[P, T](val init: Int, val finals: Set[Int], val transitions: Map[Int, List[(P, Int)]])(implicit
```
ybasket, line 24: Two questions come to my mind here:

1. Should it be `private` to minimise MiMa surface? Are we expecting any user of the library to work with this directly?
2. Are the `Int`s representing the states sequential? Because then we could use an `IndexedSeq` (or even `Array`) for faster access that the `Map`.

finite-state/shared/src/main/scala/fs2/data/pfsa/PNFA.scala
```diff
@@ -0,0 +1,136 @@
...
+            val newQ = newStates.size
+            val newStates1 = newStates.updated(q, newQ)
+            val ts = q.toList
+              .flatMap(transitions.get(_))
+              .flatMap(_.collect { case (Some(p), q) =>
+                (p, q)
+              })
+              .groupMap(_._1)(_._2)
+              .view
+              .mapValues(epsilonClosure(_))
```
ybasket, line 95:
```suggestion
              .fmap(epsilonClosure(_))
```

ybasket (commented):
Looks good to me, though I have to admit I can't afford spending the days I would need to properly find logic bugs in there (if there are any at all). Left two questions on code structure and performance that I'd like to see answered in any way before we merge.

PR #372, opened 2022-09-03
Add possibility to call transformation on self

satabin:
This adds a new possiblity in addition to the ability to call on
children and siblings, by allowing more powerful transformations applied
to the root itself.

### DIFF
--- a/finite-state/shared/src/main/scala/fs2/data/mft/MFT.scala
+++ b/finite-state/shared/src/main/scala/fs2/data/mft/MFT.scala
@@ -25,14 +25,9 @@ import cats.syntax.all._
 
 import scala.collection.compat._
 
-sealed trait Forest {
-  def fold[T](children: => T)(siblings: => T): T =
-    this match {
-      case Forest.First  => children
-      case Forest.Second => siblings
-    }
-}
+sealed trait Forest
 object Forest {
+  case object Self extends Forest
   case object First extends Forest
   case object Second extends Forest
 }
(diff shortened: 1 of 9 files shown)


finite-state/shared/src/main/scala/fs2/data/mft/MFT.scala
```diff
@@ -65,7 +66,7 @@ private[data] class MFT[InTag, OutTag](init: Int, rules: Map[Int, Rules[InTag, O
     * The generated ESP contains one decision tree encoding all the patterns
     * of this MFT.
     */
-  def esp[F[_]](implicit F: Sync[F]): F[ESP[F, InTag, OutTag]] = {
+  def esp[F[_]](implicit F: MonadError[F, Throwable] with Defer[F]): F[ESP[F, InTag, OutTag]] = {
```
ybasket, line 69:
```suggestion
  def esp[F[_]](implicit F: MonadError[F, Throwable], defer: Defer[F]): F[ESP[F, InTag, OutTag]] = {
```

I would not rely on both being implemented in the same place.

finite-state/shared/src/main/scala/fs2/data/mft/MFTBuilder.scala
```diff
@@ -0,0 +1,58 @@
...
+
+class MFTBuilder[InTag, OutTag] private[mft] {
+  self =>
+
+  private var initial = 0
+
+  private val states = new ListBuffer[StateBuilder]
+
+  final class StateBuilder private[mft] (val q: Int, val nargs: Int) {
+    var rules = new ListBuffer[(EventSelector[InTag], Rhs[OutTag])]
```
ybasket, line 13: Why do we need a `var` of a mutable collection here?

finite-state/shared/src/main/scala/fs2/data/mft/MFTBuilder.scala
```diff
@@ -0,0 +1,58 @@
...
+        case PatternBuilder.Leaf(in) => q.rules += (EventSelector.Leaf(in) -> rhs)
+        case PatternBuilder.AnyLeaf  => q.rules += (EventSelector.AnyLeaf -> rhs)
+        case PatternBuilder.Epsilon  => q.rules += (EventSelector.Epsilon -> rhs)
+        case PatternBuilder.Any =>
+          q.rules += (EventSelector.AnyNode -> rhs) += (EventSelector.AnyLeaf -> rhs) += (EventSelector.Epsilon -> rhs)
+      }
+
+  }
+
+  def newState(nargs: Int = 0, initial: Boolean): StateBuilder = {
```
ybasket, line 46: Let's move this to `StateBuilder` itself, as constructor or `apply` on the companion?

finite-state/shared/src/main/scala/fs2/data/pattern/Compiler.scala
```diff
@@ -16,12 +16,12 @@
...
 package fs2.data.pattern
 
-import cats.effect.Sync
+import cats.{Defer, MonadError}
 import cats.syntax.all._
 
 class PatternException(msg: String) extends Exception(msg)
 
-class Compiler[F[_], Tag, Pat, Out](heuristic: Heuristic[Tag])(implicit F: Sync[F]) {
+class Compiler[F[_], Tag, Pat, Out](heuristic: Heuristic[Tag])(implicit F: MonadError[F, Throwable] with Defer[F]) {
```
ybasket, line 24:
```suggestion
class Compiler[F[_], Tag, Pat, Out](heuristic: Heuristic[Tag])(implicit F: MonadError[F, Throwable], defer: Defer[F]) {
```

Same as above.

finite-state/shared/src/main/scala/fs2/data/mft/package.scala
```diff
@@ -0,0 +1,60 @@
+package fs2.data
+
+package object mft {
+
+  def dsl[InTag, OutTag](build: MFTBuilder[InTag, OutTag] => Unit): MFT[InTag, OutTag] = {
```
ybasket, line 5: Would love to see an overload for Scala 3 that takes a context function.

finite-state/shared/src/main/scala/fs2/data/mft/MFTBuilder.scala
satabin: Indeed this is leftover from previous attempt.

finite-state/shared/src/main/scala/fs2/data/mft/package.scala
satabin: :grimacing: OK

PR #384, opened 2022-09-19
Compile most modules with Scala Native

### LINKED ISSUE #376: Scala Native
Weaver snapshots are available:
- https://github.com/disneystreaming/weaver-test/issues/564

ybasket:
Had to adjust some Java Time related tests slightly as we don't have a proper timezone DB on native, but other than that, it was very easy. ~Diffson needs to be cross-compiled first to unblock the other modules, looking into that.~

Closes #376.


ybasket:
Failing on Scala 3 because Play! JSON isn't available there yet via Diffson (to be fixed with https://github.com/gnieh/diffson/pull/303), works with locally published build.

Other than that, this PR is close to being ready.

build.sbt
```diff
@@ -140,31 +155,45 @@ val root = (project in file("."))
...
     jsonCirce.js,
+    jsonCirce.native,
     jsonDiffson.jvm,
     jsonDiffson.js,
+    jsonDiffson.native,
     jsonInterpolators.jvm,
     jsonInterpolators.js,
+    jsonInterpolators.native,
+    jsonPlay.jvm,
+    jsonPlay.js,
```
ybasket, line 178: The fact they were missing didn't seem to bother anyone, the only published version was 1.3.0: https://mvnrepository.com/artifact/org.gnieh/fs2-data-json-play

Trying to fix this now within this PR by releasing full Diffson support.

I'm in favour of dropping support in v2, that library seems rather dead and the code we maintain is so small that it could easily be inlined in users' projects.

csv/shared/src/main/scala-2.13+/fs2/data/csv/LiteralCellDecoders.scala
```diff
@@ -18,10 +18,9 @@ package fs2.data.csv
```
ybasket, line 18: Changes to fix compiler warnings from 3.2.0.

PR #385, opened 2022-10-04
Fix XML rendering

ybasket:
Fix buggy Show instance for XmlEvent and provide a pipe to render events with collapsed empty tags.

### DIFF
--- a/xml/src/main/scala/fs2/data/xml/package.scala
+++ b/xml/src/main/scala/fs2/data/xml/package.scala
@@ -75,6 +75,18 @@ package object xml {
   def normalize[F[_]]: Pipe[F, XmlEvent, XmlEvent] =
     Normalizer.pipe[F]
 
+  /**
+    * Render the incoming xml events to their string representation. The output will be concise,
+    * without additional (or original) whitespace and with empty tags being collapsed to the short self-closed form <x/>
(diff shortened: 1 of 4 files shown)


satabin (approved):

xml/src/main/scala/fs2/data/xml/package.scala
```diff
@@ -75,6 +75,18 @@ package object xml {
...
 
+  /**
+    * Render the incoming xml events to their string representation. The output will be concise,
+    * without additional (or original) whitespace and with empty tags being collapsed to the short self-closed form <x/>
+    * if collapseEmpty is true. Preserves chunking, each String in the output will correspond to one event in the input.
+    */
+  def render[F[_]](collapseEmpty: Boolean = true): Pipe[F, XmlEvent, String] =
+    _.zipWithPrevious.map {
+      case (_, st: XmlEvent.StartTag)                                                 => st.render(collapseEmpty)
+      case (Some(XmlEvent.StartTag(_, _, true)), XmlEvent.EndTag(_)) if collapseEmpty => ""
```
ybasket, line 86: @satabin Just confirming: It is correct that we emit an `EndTag` after a `StartTag` with `isEmpty=true`, right?

So `<doc/>` would lead to `Lis(StartTag("doc", Nil, true), EndTag("doc"))`.

PR #315, opened 2022-04-16
Switch to sbt-typelevel

satabin:

PR #388, opened 2022-10-05
Remove setup-gpg, refresh release workflow

ybasket:
Hopefully fixes the problem we have with releasing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,13 +9,12 @@ on:
       - "v*"
 jobs:
   publish:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2.3.4
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - uses: olafurpg/setup-scala@v10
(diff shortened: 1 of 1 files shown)


satabin (approved):

.github/workflows/release.yml
```diff
@@ -9,13 +9,12 @@ on:
...
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2.3.4
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - uses: olafurpg/setup-scala@v10
-      - uses: olafurpg/setup-gpg@v3
+      - uses: olafurpg/setup-scala@v13
```
keynmol, line 17: Btw, Olaf's setup-scala is being deprecated.
I'd recommend using builtin `actions/setup-java` (explicitly setting java version to whatever your target is) which already comes with both sbt and sbt cache configuration

ybasket: Thanks for the hint! I'm aware of the deprecation, but don't want to change much on what was supposed to be a "simple" patch release. #315 would be something I want to get over the finish line, especially now with the native support included.

PR #390, opened 2022-10-06
Add clarification for separator in high level API

zetashift:
Better late than never :D.
I'm not sure if I put the clarification in the right place, open to any kind of feedback.


ybasket:
Looks good to me. Thank you for sending this PR!

PR #393, opened 2022-10-07
Use the org secret for PGP key

satabin:


ybasket (approved):

ybasket:
If we do another 1.5.x release, we need to remember to cherry-pick this or re-add the old secret temporarily.

satabin:
I didn't delete the old one yet.

PR #394, opened 2022-10-07
Add guard to pattern matching

satabin:
Pattern matching with guards comes in handy to check values instead of structure.

This will be particularly useful to implement XPath predicates in the XQuery implementation.

Implementation based on [_Compiling Successor ML Pattern Guards_](https://people.cs.uchicago.edu/~jhr/papers/2019/ml-guards.pdf)


finite-state/shared/src/main/scala/fs2/data/esp/ESP.scala
```diff
@@ -109,38 +107,28 @@ private[data] class ESP[F[_], InTag, OutTag](init: Int,
...
-          .liftTo[Pull[F, Nothing, *]](new ESPException(s"unknown captured input $name"))
-          .flatMap { selector =>
-            in.flatMap(selectInTag(_, selector))
-              .liftTo[Pull[F, Nothing, *]](new ESPException("cannot caputre eos"))
-              .map { v =>
-                val tag = TT.convert(v)
-                Expr.Leaf(Out.makeLeaf(tag), Expr.Epsilon)
-              }
+        in.flatMap(select(_, Selector.Cons(Selector.Root, Tag.Leaf, 0)))
+          .liftTo[Pull[F, Nothing, *]](new ESPException("cannot caputre eos"))
```
ybasket, line 127:
```suggestion
          .liftTo[Pull[F, Nothing, *]](new ESPException("cannot capture eos"))
```

ybasket, line 127:
```suggestion
          .liftTo[Pull[F, Nothing, *]](new ESPException("cannot capture eos"))
```

finite-state/shared/src/main/scala/fs2/data/esp/ESP.scala
```diff
@@ -109,38 +107,28 @@ private[data] class ESP[F[_], InTag, OutTag](init: Int,
...
+          .map { v =>
+            val tag = TT.convert(v)
+            Expr.Leaf(Out.makeLeaf(tag), Expr.Epsilon)
           }
```
ybasket, line 131:
```suggestion
          .map(Expr.Leaf(Out.makeLeaf(TT.convert(v)), Expr.Epsilon))
```

finite-state/shared/src/main/scala/fs2/data/pattern/DecisionTree.scala
```diff
@@ -39,7 +37,10 @@ sealed trait DecisionTree[Tag, Out] {
                   }
               }
             case None =>
-              None
+              fallBack match {
+                case Some(tree) => loop(tree)
+                case None       => None
+              }
```
ybasket, line 43: Isn't that effectively a `flatMap`?

```suggestion
              fallBack.flatMap(loop)
```

satabin: Yes it is, but `@tailrec` optimization is here in my implementation :sweat_smile:

ybasket: OK, I let that argument count 😄

ybasket (approved):
Looks good (I believe, don't feel like I could spot an error in the algorithm in this review)! Left a couple of style comments, feel free to pick/ignore as you see fit and then merge.

finite-state/shared/src/test/scala/fs2/data/pattern/PatGuardSpec.scala
```diff
@@ -0,0 +1,81 @@
...
+      )
+
+      compiler.compile(cases)
+    }
+
+  implicit val showText: Show[MiniXML.Text] = Show.show(_.txt)
+
+  test("Guard 'a'") { tree =>
+    IO(expect.same(Some(1), tree.get(Input(0, 0, MiniXML.open("a").some))))
+  }
```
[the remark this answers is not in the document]
satabin: Nope, there is not overload of `pureTest` that takes the resource as input.

PR #399, opened 2022-10-11
Make guard type abstract and integrate in MFT

satabin:
Not all rewriting languages need guards, and guards could be different from format to format. The goal of this PR is to abstract over the guard type, and providing the proper instances when no guard exists in the language.


finite-state/shared/src/main/scala/fs2/data/esp/Pattern.scala
```diff
@@ -25,153 +25,142 @@ import cats.syntax.all._
 /** A pattern to be matched.
   * A pattern can capture a submatch, that is then usable in the RHS.
   */
-sealed trait Pattern[+Tag] {
-  def |[Tag1 >: Tag](that: Pattern[Tag1]): Pattern[Tag1] =
+sealed trait Pattern[Guard, Tag] {
```
ybasket, line 28: Any specific reason you removed the covariance on `Tag`?

finite-state/shared/src/main/scala/fs2/data/esp/Guard.scala
```diff
@@ -1,104 +0,0 @@
-/*
```
ybasket, line 1: Was it not worth to keep this a possible implementation for guards even though it's no longer the exclusive one? Seemed general enough to be useful several times.

ybasket (approved):
I'm too far from this part of the code base to really judge whether it's a good change. Conceptually it sounds good, but the diff itself is mostly adding complexity and removing bits that seemed useful. I'll see once XPath becomes reality what this is all about.

So take my two questions as hints what be sure about and go ahead as you see fit.

finite-state/shared/src/main/scala/fs2/data/esp/Pattern.scala
satabin: To make it compile with scala 2.12 and scala 3, for which type inference in patterns for types with 2 covariant parameters seems broken.

finite-state/shared/src/main/scala/fs2/data/esp/Guard.scala
satabin: Actually the `In` and `NotIn` operators and their evaluation was terribly specific and redundant with the XPath predicate that we have.

ybasket: OK then.

PR #436, opened 2023-01-11
Adds attempt decoding functions

alycklama:
@GerretS and I worked on a potential improvement regarding errors as values. We noticed that when parsing CSV files, the default behaviour is to fail the stream, whenever an error occurs. For our use case, we would like to aggregate these errors at the end, so that we have all potential errors at the same time.

There are a few open questions that we would like your input on.

**Tests have been added to the csv-generic**
As we wanted to test the decoding of a case class, we added the tests for all our `attempt*` functions to the `csv-generic` project. Is that OK, or would you like us to change that. We didn't want to manually create our own decoder within `csv/shared`.

**headersAttempt (bincompat)**
The existence of the headersAttempt bincompat forces us to pass the implicits along, instead of them implicitly being passed along. Is that OK for now? The only solution that we saw is to remove the bincompat version, but we have no idea what the side effects are there.

**Verbose tests**
The tests are now a bit verbose, as there's quite some duplication. However, the tests are rather trivial and we wonder if you are OK with the setup, or whether we should improve that.

We're looking forward hearing from you!


alycklama:
We noticed that there's a difference between Scala 2 and Scala 3. One of the tests now fails unexpectedly.

```
    case class TestData(name: String, age: String, description: String)

    val content =
      """name1,age,description
        |John Doe,47,description 1
        |Bob Smith,80,description 2
        |""".stripMargin
```

In Scala 3, decoding the CSV into the `TestData` case class will not throw any exception. Instead, the `name1` column is ignored, resulting in an empty string value for the `name` variable in the `TestData` case class.

In Scala 2, decoding the CSV into the `TestData` case class will result in the error "unknown column name 'name' in line x".

Any idea if this is intentional?

ybasket:
> We noticed that there's a difference between Scala 2 and Scala 3. One of the tests now fails unexpectedly.
> 
> ```
>     case class TestData(name: String, age: String, description: String)
> 
>     val content =
>       """name1,age,description
>         |John Doe,47,description 1
>         |Bob Smith,80,description 2
>         |""".stripMargin
> ```
> 
> In Scala 3, decoding the CSV into the `TestData` case class will not throw any exception. Instead, the `name1` column is ignored, resulting in an empty string value for the `name` variable in the `TestData` case class.
> 
> In Scala 2, decoding the CSV into the `TestData` case class will result in the error "unknown column name 'name' in line x".
> 
> Any idea if this is intentional?

Both Scala versions are meant to behave consistent.

Thank you for your PR! I don't yet have the time to look at it closely (should find some the next days though), but will approve the workflow runs if you push more stuff so you get CI feedback until we can properly review it.

csv/shared/src/main/scala/fs2/data/csv/package.scala
```diff
@@ -162,13 +220,42 @@ package object csv {
         C: CharLikeChunks[F, C],
         T: CsvRowDecoder[T, Header]): Pipe[F, C, T] = {
       if (skipHeaders)
-        lowlevel.rows(separator, quoteHandling) andThen (_.tail) andThen
+        lowlevel.rows(separator, quoteHandling) andThen lowlevel.skipHeaders andThen
```
ybasket, line 223: This doesn't actually make a difference (same implementation) though intent becomes clearer. You stated in your questions something about empty stream failing, can you elaborate? A `.tail` on an empty stream will just be an empty stream again, so no error, no problem.

csv/generic/shared/src/test/scala/fs2/data/csv/generic/CsvRowDecoderTest.scala
```diff
@@ -18,9 +18,11 @@ package fs2
 package data.csv
 package generic
 
+import cats.Eq
 import semiauto._
 import cats.data.NonEmptyList
-
+import cats.effect.IO
+import cats.implicits.catsSyntaxEq
```
ybasket, line 25:
```suggestion
import cats.syntax.eq._
```

csv/generic/shared/src/test/scala/fs2/data/csv/generic/CsvRowDecoderTest.scala
```diff
@@ -82,4 +87,161 @@ object CsvRowDecoderTest extends SimpleIOSuite {
       expect(testOptionalStringDecoder(csvRowEmptyCell) == Right(TestOptionalString(1, None, 42)))
   }
 
+  implicit val decoderErrorEq: Eq[Throwable] = (a, b) => a.toString === b.toString
+
+  implicit val decoderTestDataEq: Eq[TestData] = (a, b) => a == b
```
ybasket, line 92:
```suggestion
  implicit val decoderTestDataEq: Eq[TestData] = Eq.fromUniversalEquals
```

csv/generic/shared/src/test/scala/fs2/data/csv/generic/CsvRowDecoderTest.scala
```diff
@@ -82,4 +87,161 @@ object CsvRowDecoderTest extends SimpleIOSuite {
...
+    val expected: List[Either[Throwable, TestData]] = List(
+      Right(TestData("John Doe", "47", "description 1")),
+      Left(new CsvException("Headers have size 3 but row has size 2. Both numbers must match! in line 3", None)),
+      Right(TestData("Bob Smith", "80", "description 2")),
+      Left(new CsvException("Headers have size 3 but row has size 2. Both numbers must match! in line 5", None))
+    )
+
+    Stream
+      .emit(content)
+      .covary[IO]
```
ybasket, line 112: You don't need `IO` in these tests, `Fallible` will do just fine as we don't have to read any files.

ybasket (commented):
Is your use case to parse CSV files that have rows with a wrong numbers of columns in between or just erroneous data (as in: no number where expected and alike)?

I'm not entirely convinced yet that these combinators are justified on the high level API, as they are easy to do using the low level API and seem to be rather non-standard use cases. But happy to learn more!

As for the implicits, that's fine. Maintaining bincompat is sometimes slightly ugly and at least, it stays internal to the library and we can clean up in fs2-data 2.0 one day.

Test verbosity would be OK, but let's put them where they belong (`csv` module), maybe in a new file. I'll add some inline comments for the tests as well.

GerretS:
> Is your use case to parse CSV files that have rows with a wrong numbers of columns in between or just erroneous data (as in: no number where expected and alike)?

People sometimes create CSV files by manually inputting data into Excel or whatever, and those files become invalid in the weirdest ways. Strings in an expected Int field, commas within an unquoted String and who knows what else. In that situation, it's much more convenient to be able to return a full list of errors for a file so they can fix them all at once than only returning the first error and requiring the user to try over and over again until they fix all of them.

The use case is similar to how you could use Cats Validated to parse several sources of data in parallel and return a list of all errors instead of throwing whatever happens to be the first exception.


As for your other comments, we'll take a look after the weekend.

alycklama:
It could be a matter of perspective. When using a streaming library, especially in a functional programming language like Scala, I expect errors to be first-class citizens. So it surprised me that the default behaviour is to kill the stream whenever an error is encountered. That's not something I would have expected from a resiliency perspective. In my opinion, returning errors as values is not a novelty feature someone should implement via the low-level API. I would have expected this to be available in the high-level API, though, as I mentioned initially, it could all be a matter of perspective. It took quite a while to figure out that aggregating errors is not achievable via the high-level API and that one should use the low-level API instead. It would be great to have the documentation reflect this because it could have been more evident with a concrete example of achieving this.

ybasket:
>

GerretS:
I agree there are two different kinds of errors.

But in case of a row column not matching the header column count, this could be a single error for that row because someone accidentally deleted a comma. It would make sense to me to handle those as single-row errors and accumulate them.

alycklama:
No worries for the close/re-open!

If you're dealing with manually created CSV files, then a single row that cannot be decoded is to be expected. That should, however, not affect the entire file from a resiliency point of view.

The only way I could figure out so far is to not use `decodeUsingHeaders` and instead use `decodeSkippingHeaders`. The problem is that the order of the columns in the case class needs to be precisely the same as the CSV file. The problem of having the columns ordered in the exact same way as the case class representation also holds for the low-level API.

Suppose we can get the headers per row by having an `attemptUsingHeaders`. The ordering can differ between the CSV and the case class representation in that case.

How would you create a resilient stream that does not fail, even if the structure is incorrect (e.g. a column is missing for a row) using the low-level API?

```
    import fs2._
    import fs2.data.csv._
    import fs2.data.csv.generic.semiauto._

    case class Data(name: String, description: String)
    implicit val testDataRowDecoder: RowDecoder[Data] = deriveRowDecoder[Data]
    implicit val testCsvDataRowDecoder: CsvRowDecoder[Data, String] = deriveCsvRowDecoder[Data]

    val csv =
      """description,name
        |hello world,John
        |this is a test
        |""".stripMargin

    val results1 = Stream
      .emit(csv)
      .covary[Fallible]
      .through(decodeSkippingHeaders[DecoderResult[Data]]())
      .compile
      .toList

    val results2 = Stream
      .emit(csv)
      .covary[Fallible]
      .through(decodeUsingHeaders[DecoderResult[Data]]())
      .compile
      .toList

    val results3 = Stream.emit(csv).through(lowlevel.rows[Fallible, String](',', QuoteHandling.Literal))
      .map(testDataRowDecoder.apply)
      .compile
      .toList

    // Using decodeSkippingHeaders:
    // results1: Right(List(Right(Data(hello world,John)), Left(fs2.data.csv.DecoderError: unexpect end of row in line 3)))
    //
    // Using decodeUsingHeaders:
    // results2: Left(fs2.data.csv.CsvException: Headers have size 2 but row has size 1. Both numbers must match! in line 3)
    //
    // Using the low-level API:
    // result3: Right(List(Right(Data(description,name)), Right(Data(hello world,John)), Left(fs2.data.csv.DecoderError: unexpect end of row in line 3)))
```

csv/generic/shared/src/test/scala/fs2/data/csv/generic/CsvRowDecoderTest.scala
```diff
@@ -82,4 +87,161 @@ object CsvRowDecoderTest extends SimpleIOSuite {
       expect(testOptionalStringDecoder(csvRowEmptyCell) == Right(TestOptionalString(1, None, 42)))
   }
 
+  implicit val decoderErrorEq: Eq[Throwable] = (a, b) => a.toString === b.toString
```
satabin, line 90:
```suggestion
  implicit val decoderErrorEq: Eq[Throwable] = Eq.by(_.toString)
```

ybasket:
> How would you create a resilient stream that does not fail, even if the structure is incorrect (e.g. a column is missing for a row) using the low-level API?
Same as you do in your implementations in this PR.

I discussed this with @satabin and we agreed that because varying column count is not generally safe to parse per row (because it can well be a sign of the whole file being corrupted), we would like to namespace the new pipes into a `lenient` object. That way there's a clear signal (and hopefully documentation) that users need to think carefully about their use case before using them. 

So if you could move the new pipes into such object (similar to the `lowlevel` one) and fix the tests as discussed above, we'd be good to go :)

csv/shared/src/main/scala/fs2/data/csv/package.scala
alycklama: In retrospective, I think the fail had to do with the headers not being able to be decoded due to column mismatches.

alycklama:
@ybasket @satabin 
I've applied your suggestions and I believe the PR to be ready. It would be great if you could have another look. Thank you in advance!

alycklama:
I've also added a bit of documentation FYI.

documentation/docs/index.md
```diff
@@ -81,6 +81,33 @@ Files[IO]
   .drain
 ```
 
+### Dealing with erroneous files
+
+The default behaviour when parsing CSV files, is to terminate the stream whenever the columns of a row do not match 
+with the columns of the header. If you're dealing with CSV files that could contain these kind of errors, you can make 
+use of the `lenient` package. You will get back a `List` of results, where each parsed row is represented by an
```
ybasket, line 88:
```suggestion
use of the `lenient` package. You will get back a `Stream` of results, where each parsed row is represented by an 
```

It's only a `List` because you use `toList` in the example, most real use cases will actually transform the results within a stream to stay in constant memory world.

csv/shared/src/main/scala/fs2/data/csv/package.scala
```diff
@@ -342,4 +358,89 @@ package object csv {
...
+      new PartiallyAppliedDecodeAttemptGivenHeaders[T](dummy = true)
+
+    class PartiallyAppliedDecodeAttemptGivenHeaders[T](val dummy: Boolean) extends AnyVal {
+      def apply[F[_], C, Header](headers: NonEmptyList[Header],
+                                 skipHeaders: Boolean = false,
+                                 separator: Char = ',',
+                                 quoteHandling: QuoteHandling = QuoteHandling.RFCCompliant)(implicit
+          F: RaiseThrowable[F],
+          C: CharLikeChunks[F, C],
+          T: CsvRowDecoder[T, Header]): Pipe[F, C, Either[Throwable, T]] = {
```
ybasket, line 383: We can narrow these down from `Throwable` to `CsvException`, can't we? That would signal the behaviour better and allow clearer error handling.

csv/generic/shared/src/test/scala/fs2/data/csv/generic/CsvRowDecoderTest.scala
```diff
@@ -18,6 +18,7 @@ package fs2
 package data.csv
```
ybasket, line 18: I would still like to see these test cases moved out of the `generic` module please, they just don't belong here.

csv/generic/shared/src/test/scala/fs2/data/csv/generic/CsvRowDecoderTest.scala
```diff
@@ -90,4 +97,446 @@ object CsvRowDecoderTest extends SimpleIOSuite {
...
+
+    stream match {
+      case Right(actual) => expect.eql(expected, actual)
+      case Left(x)       => failure(s"Stream failed with value $x")
+    }
+  }
+
+  pureTest("decodeUsingHeaders should map data to case class matching the header name to the case class field name") {
+    implicit val decoder: CsvRowDecoder[Test, String] = testDecoder
+    implicit val testEq: Eq[Test] = Eq.fromUniversalEquals
```
ybasket, line 280: Instead of defining it several times (differently) in the test cases, let's move it outside to the other new implicits.

csv/generic/shared/src/test/scala/fs2/data/csv/generic/CsvRowDecoderTest.scala
```diff
@@ -90,4 +97,446 @@ object CsvRowDecoderTest extends SimpleIOSuite {
...
+      .compile
+      .toList
+
+    stream match {
+      case Right(actual) => expect.eql(expected, actual)
+      case Left(x)       => failure(s"Stream failed with value $x")
+    }
+  }
+
+  pureTest("decodeUsingHeaders should succeed if a required field with default value is missing") {
```
ybasket, line 330: This one should fail on Scala 3 due to the missing default support, so let's remove it (for now).

ybasket (commented):
Thank you for the docs! I left some comments to be resolved still, but we're getting closer to merging 👍

csv/shared/src/main/scala/fs2/data/csv/package.scala
alycklama: Yes we can! I had to change a few things in the existing API, as there it was already a `Throwable`. See new commits for the changes.

csv/generic/shared/src/test/scala/fs2/data/csv/generic/CsvRowDecoderTest.scala
alycklama: Done! We used the decoder derive methods, but now I've added a custom decoder, as this method is no longer available in `csvJVM`. Please let me know if this is also what you had in mind.

documentation/docs/index.md
```diff
@@ -81,6 +81,33 @@ Files[IO]
...
+    |""".stripMargin
+
+Stream
+      .emit(content)
+      .covary[Fallible]
+      .through(lenient.attemptDecodeGivenHeaders[TestData]())
+      .compile
+      .toList
+
+// returns Either[Throwable, List[Either[Throwable, TestData]]]
```
satabin, line 107: This comment should not be here, I'd like to see the mdoc output (hence removing the `silent`) in the documentation.

satabin: I just realized this documentation belongs to the CSV documentation, not the root one. Can you please move it to [this file](https://github.com/gnieh/fs2-data/blob/main/documentation/docs/csv/index.md)?

alycklama: Good catch! I've moved it to the appropriate location and added a bit more content.

satabin: Perfect, thanks. I'm fine if @ybasket is.

satabin (approved):

alycklama:
@ybasket I've requested a review from you, just to be sure that we're not waiting on each other. If there's something still to be done in this PR, please let me know! 👍

ybasket:
> @ybasket I've requested a review from you, just to be sure that we're not waiting on each other. If there's something still to be done in this PR, please let me know! 👍

👍 I was on vacation the last days, so sorry for not replying earlier!

Code looks good, I'm just wondering why there are still changes in the `generic` module – they don't seem to be specific to `generic` and covered by other tests, so let's remove or move them? Once that's resolved I'm happy to merge! 

Also saw during review we have a specific exception type (`<: CsvException`) for misaligned rows, but weren't using it, so I would do a follow-up PR and then release a new version with all CSV fixes.

alycklama:
@ybasket No problem! Hope you had nice days off.

I've removed the tests in the generic module.

> Also saw during review we have a specific exception type (`<: CsvException`) for misaligned rows, but weren't using it, so I would do a follow-up PR and then release a new version with all CSV fixes.

What do you mean here? All the `Throwable` types have been replaced by `CsvException` as that's the supertype of the potential errors that can be thrown. What kind of follow-up PR do you expect? And is this something you expect us to do, or is this something you will pick-up?

ybasket:
I mean the specialised error type here: https://github.com/gnieh/fs2-data/blob/f2dd6e392fa765af9fbc139b6c34b04751232356/csv/shared/src/main/scala/fs2/data/csv/exceptions.scala#L35

I'll do the follow-up (it was apparently my oversight long ago), no worries. It's also compatible with your changes. And I'll also clean up the small remainders of changes in the `generic` module that are still in this PR and merge. Thank you for your work 😃 @alycklama @GerretS

alycklama:
You're welcome! Thank you in advance for approving and merging the PR! 😄

PR #453, opened 2023-02-26
Improve JSON Parser Performances

satabin:
Follow-up from #489 

This PR implements several optimizations, resulting from observations from profilers:
 - Smarter `CharLikeChunk` implementations
   - It avoids adding characters one by one in the result and offers a way to batch copy a range in a new buffer
   - It accesses characters in an array instead of a string, as `String.charAt` makes more checks than array access 
   - Since the new ones are not binary compatible, the parser specializes for the new implementations under the hood
   - Inheritance is marked as deprecated and trait will be sealed in the future
   - Other text based parser will be able to benefit from the improvements
 - Implementation of a faster stateless version of string parsing when there is no escape and fall back to the slower one on the first escape encountered in the string
 - Make various buffer size configurable
   - Depending on the specific data users are handling, they might want to allocate bigger or smaller buffer when building strings for keys, string values, or numbers.
   - The right initial buffer size can avoid buffer resizing (and hence data copy)
   - No documentation is added for those settings yet as they are experimental and might change in the future
 - Delete various boxing and `Option` allocations throughout the parser.
 - The improvements performed in this PR do not change the way the AST is built. This work will be done in a following PR.

With these improvements in place, we get these results on my machine (openjdk 19, Nixos, Intel Core i5-6500, 24GiB RAM):
```[info] Benchmark                                    Mode  Cnt     Score    Error  Units
JsonParserBenchmarks.parseCirceFs2           avgt   10  5541.715 ± 24.544  us/op
JsonParserBenchmarks.parseJsonFs2DataTokens  avgt   10  3800.039 ± 46.111  us/op
JsonParserBenchmarks.parseJsonFs2DataValues  avgt   10  6716.085 ± 60.833  us/op
JsonParserBenchmarks.parseJsonJawn           avgt   10  2019.646 ± 41.275  us/op
JsonValueBenchmarks.parseEscapedString       avgt   10     2.996 ±  0.069  us/op
JsonValueBenchmarks.parseSimpleString        avgt   10     2.712 ±  0.063  us/op
```
Compared to jawn the token parser is now only ~1.88x slower (previously ~2.57x slower) and the value parser is ~3.33x slower (previously ~3.85x slower).

Some progress can probably be achieved in the AST construction and will be addressed later, to bring fs2-data on par with circe-fs2.


text/shared/src/main/scala/fs2/data/text/CharLikeChunks.scala
```diff
@@ -69,103 +74,112 @@ trait CharLikeChunks[F[_], In] {
...
-  def needsPull(ctx: Context): Boolean =
-    ctx.idx >= ctx.chunk.size
+private[text] class CharLikeSingleByteChunks[F[_]](charset: Charset) extends CharArrayBuffer[F, Byte] {
 
   def pullNext(ctx: Context): Pull[F, Nothing, Option[Context]] =
     ctx.rest.pull.uncons.map(_.map { case (hd, tl) =>
       // This code is pure. The constructor replaces malformed input
       // with the charset default replacement string, so this never throws
-      ctx.chunk = new String(hd.toArray[Byte], charset)
+      ctx.chunk = new String(hd.toArray[Byte], charset).toCharArray()
```
ybasket, line 148: Would require benchmarking, but I could imagine that `charset.decode(hd.toByteBuffer).array()` is faster. Or maybe one even wants to replace `Array[Char]` with `CharBuffer`?

ybasket (approved):
Left one comment, but looks good, nice work!

text/shared/src/main/scala/fs2/data/text/CharLikeChunks.scala
satabin: I think I will eventually switch to `CharBuffer`, I don't know why I didn't yet. I will give it a try.

satabin: After trying a bit various things it requires more investigations on my side. I will merge this PR, which lays out the ground work, and will open another one with `CharBuffer` if it is worth it. Thanks for pointing this out.

PR #454, opened 2023-02-26
Decode bignums properly

### LINKED ISSUE #452: CBOR doesn't round trip bignums
When encoding large integers to bytes and back, the CBOR (de)serialisation doesn't produce the same as the input as it handles bignum tags only when encoding, not on the way back:

```scala
import cats.effect.SyncIO
import fs2.*
import fs2.data.cbor.*
import fs2.data.cbor.high.*

val value = CborValue.Integer(BigInt("-739421513997118914047232662242593364"))

val out =  Stream(value).covary[SyncIO].through(toBinary).through(values).compile.onlyOrError.unsafeRunSync()

println(value) // Integer(-739421513997118914047232662242593364)
println(out) // Tagged(3,ByteString(ByteVector(16 bytes, 0x008e684b9913062199945815948c0a53)))
```

Scastie: https://scastie.scala-lang.org/VuVAwL2wQiqrcWiSeyCIeg

ybasket:
Add support for bignum decoding to match the encoding logic for high-level integer values and hence enable roundtrip for them.

Closes #452

### DIFF
--- a/cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
+++ b/cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
@@ -219,20 +219,20 @@ object ValueParser {
     } else {
       chunk(idx) match {
         case CborItem.Tag(Tags.PositiveBigNum) =>
-          def f: CborValue => Pull[F, Nothing, CborValue] = {
-            case CborValue.ByteString(bytes) => tags(CborValue.Integer(BigInt(bytes.toArrayUnsafe)))
-            case CborValue.Tagged(tag2, v)   => f(v).map(CborValue.Tagged(tag2, _))
+          def f(recurse: Boolean): CborValue => Pull[F, Nothing, CborValue] = {
+            case CborValue.ByteString(bytes)          => tags(CborValue.Integer(BigInt(bytes.toArrayUnsafe)))
+            case CborValue.Tagged(tag2, v) if recurse => tags(CborValue.Tagged(tag2, v)).flatMap(f(false))
             case _ => Pull.raiseError(new CborTagDecodingException("A bignum must have a byte string as value"))
           }
-          parseTags(chunk, idx + 1, rest, f, chunkAcc)
+          parseTags(chunk, idx + 1, rest, f(true), chunkAcc)
         case CborItem.Tag(Tags.NegativeBigNum) =>
-          def f: CborValue => Pull[F, Nothing, CborValue] = {
+          def f(recurse: Boolean): CborValue => Pull[F, Nothing, CborValue] = {
             case CborValue.ByteString(bytes) =>
               tags(CborValue.Integer(minusOne + minusOne * BigInt(bytes.toArrayUnsafe)))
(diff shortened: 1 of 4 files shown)


cbor/shared/src/test/scala/fs2/data/cbor/high/ValueParserTest.scala
```diff
@@ -0,0 +1,16 @@
+package fs2.data.cbor.high
+
+import cats.effect.IO
+import fs2._
+import weaver.SimpleIOSuite
+
+object ValueParserTest extends SimpleIOSuite {
```
ybasket, line 7: I started some Scalacheck-based roundtrip testing in http4s-fs2-data, but maybe it's better done here. I'll port it over or at least extend this test to capture positive bigints as well.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
```diff
@@ -209,16 +208,31 @@ object ValueParser {
...
       Pull.output(Chunk.seq(chunkAcc.reverse)) >> rest.pull.uncons.flatMap {
         case Some((hd, tl)) => parseTags(hd, 0, tl, tags, Nil)
         case None           => Pull.raiseError(new CborParsingException("unexpected end of input"))
       }
     } else {
       chunk(idx) match {
+        case CborItem.Tag(Tags.PositiveBigNum) =>
+          def f: CborValue => Pull[F, Nothing, CborValue] = {
+            case CborValue.ByteString(bytes) => tags(CborValue.Integer(BigInt(bytes.toArrayUnsafe)))
+            case CborValue.Tagged(tag2, v)   => f(v).map(CborValue.Tagged(tag2, _))
```
satabin, line 224: I think this is not correct. The specification states:

> A tag applies semantics to the data item it encloses. Tags can nest: if tag A encloses tag B, which encloses data item C, tag A applies to the result of applying tag B on data item C.

So `f` should apply to the decoding of value tagged by `tag2`.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
```diff
@@ -209,16 +208,31 @@ object ValueParser {
...
+            case _ => Pull.raiseError(new CborTagDecodingException("A bignum must have a byte string as value"))
```
satabin, line 225: What about Indefinite [Length Byte String](https://www.rfc-editor.org/rfc/rfc8949.html#name-indefinite-length-byte-stri)? They are byte strings after all, and this would be valid as an encoding I guess.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
ybasket: Should be fixed now.

cbor/shared/src/main/scala/fs2/data/cbor/high/internal/ValueParser.scala
satabin: This is true, I misread, indeed.

satabin (approved):

satabin (approved):

PR #459, opened 2023-03-10
Generalize optional fields handling for `RowF`

satabin:
An optional field might be encoded in various ways, depending on the CSV data emitter. `RowF` can handle most scenarios by providing a way to define:
 - how to react when a field is not present in the dataset at all (i.e. the column is missing in the input data)
 - what values are considered empty (e.g. empty string, or `null`, or `N/A`)


ybasket (approved):
LGTM

PR #465, opened 2023-04-01
Fix CBOR number encoding

ybasket:
Serialisation of CBOR numbers missed byte padding for 32 and 64-bit wide numbers, so they were not encoded in the full length and destroyed the validity of the whole stream.

### DIFF
--- a/cbor/shared/src/test/scala/fs2/data/cbor/RoundtripTest.scala
+++ b/cbor/shared/src/test/scala/fs2/data/cbor/RoundtripTest.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2023 Lucas Satabin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


cbor/shared/src/test/scala/fs2/data/cbor/RoundtripTest.scala
```diff
@@ -0,0 +1,26 @@
...
+package data
+package cbor
+
+import cats.Show
+import fs2.data.cbor.high.CborValue
+import org.scalacheck.Arbitrary
+import weaver.SimpleIOSuite
+import weaver.scalacheck.Checkers
+
+object RoundtripTest extends SimpleIOSuite with Checkers {
```
ybasket, line 11: We should extend this kind of test to all types of CBOR values, I did this in http4s-fs2-data and found two bugs already. Left for a follow-up.

satabin (approved):

PR #491, opened 2023-06-26
Bring JSON parser on par with circe-fs2

satabin:
The goal of this PR is to have performances on par with circe-fs2, so that users can switch fearlessly to fs2-data (see circe/circe-fs2#425).
The existing two steps approach has a major drawback, that prevents this from happening: it builds intermediate `Token`, which are immediately consumed to build the AST using an instance of `Builder`.
This can be avoided for people coming from circe-fs2, by adding a new pipe `ast.parse` that directly transforms the input byte/char/string stream into the AST.

To achieve this, this PR takes inspiration from the `Facade` approach in jawn to model the `ChunkAccumulator` in the JSON parser. One such simple accumulator simply wraps the existing `VectorBuilder[Token]` and makes it possible to build the token stream. Another implementation (`BuilderChunkAccumulator[Json]`) directly builds the AST from the parser without instantiating any `Token`. In the future, it can even be extended to implement filtering at the source, without constructing the `Token`s that are not selected. The abstraction is kept internal for now, so that we have more latitude to change the API and stabilize it. 

 - [X] Implement the parser in terms of `ChunkAccumulator`
 - [X] Implement the simple `TokenChunkAccumulator`
 - [X] Implement the AST building `BuilderChunkAccumulator`
   - [X] implement `ast.parse`
 - [x] Add sclafix rule to migrate from `.through(tokens).through(ast.values)` to `.through(ast.parse)`
 - [x] Add `ast.parse` documentation
 - [x] Add cookbook to migrate from circe-fs2

Using the new abstraction, here are the results of the benchmark.

```
Benchmark                                          Mode  Cnt     Score    Error  Units
JsonParserBenchmarks.parseCirceFs2                 avgt   10  5197.451 ± 18.083  us/op
JsonParserBenchmarks.parseJsonFs2DataParse         avgt   10  5259.336 ± 32.579  us/op
JsonParserBenchmarks.parseJsonFs2DataTokens        avgt   10  3531.900 ± 15.076  us/op
JsonParserBenchmarks.parseJsonFs2DataTokensValues  avgt   10  6091.062 ± 94.753  us/op
JsonParserBenchmarks.parseJsonJawn                 avgt   10  1945.402 ± 10.970  us/op
JsonValueBenchmarks.parseEscapedString             avgt   10     2.742 ±  0.021  us/op
JsonValueBenchmarks.parseSimpleString              avgt   10     2.588 ±  0.027  us/op
```


satabin:
@ybasket this PR is ready to be reviewed!

documentation/docs/json/libraries.md
```diff
@@ -113,6 +109,38 @@ val transformed = stream.through(codec.transformOpt(f1, (i: Int) => (i > 0).guar
...
+import fs2.data.json._
+import fs2.data.json.circe._
+
+input[Fallible]
+  .through(ast.parse)
+  .map(_.spaces2)
+  .compile
+  .toList
+```
+
```
ybasket, line 143: Maybe we should add a reference to the docs about `CharLikeChunks` here so people that were using `byteStreamParser` know what to import?

json/src/main/scala/fs2/data/json/internal/JsonTokenParser.scala
```diff
@@ -210,88 +212,95 @@ private class JsonTokenParser[F[_], T](s: Stream[F, T])(implicit F: RaiseThrowab
...
+          keyword_(expected, eidx, elen, accumulate)
         case None => Pull.raiseError[F](new JsonException("unexpected end of input"))
       }
     } else {
       val c = T.current(context)
       if (c == expected.charAt(eidx)) {
         if (eidx == elen - 1) {
           T.advance(context)
-          chunkAcc += token
+          accumulate()
```
ybasket, line 243: Can you quickly add a comment on what's happening here? I would expect a `() => ChunkAccumulator[Res]` with discarded result to be a no-op…reading the call sites below I suspect this emits the keyword-equivalent value via the facade and we don't need the result as we don't pass it on, but maybe we can make this clearer, like calling it `emitKeywordValue`?

scalafix/rules/src/main/scala/fix/JsonParse.scala
```diff
@@ -0,0 +1,40 @@
+package fix
```
ybasket, line 1: I guess we should add the migration to Scala Steward when releasing so our users profit more directly, right?

ybasket (approved):
Nice work and sorry for the time it took me to review it!

scalafix/rules/src/main/scala/fix/JsonParse.scala
satabin: Yes But I need this PR to be merged, this is the plan afterwards.

json/src/main/scala/fs2/data/json/internal/JsonTokenParser.scala
satabin: I am not a big fan of this part so far, but this is the best I came up with for now. Renaming will help, indeed.

PR #516, opened 2023-09-10
Fix output duplication on escaped chunk boundaries

### LINKED ISSUE #515: tokens duplicates text if stream has chunks that end with escape character
I ran into an issue where reading in json file of filepaths would sometimes duplicate text when running it through tokens. The following code when run with the attached sample text reproduces the issue. The second token is parsed as \\pathpath\\to\some\other\file.txt instead of \\path\\to\some\other\file.txt with chunkSize of 55

    val chunkSize = 55
    val stream: fs2.Stream[IO, Byte] = fs2.io.readInputStream(IO(getClass.getResourceAsStream("/Sample.txt")), chunkSize)
    stream.map(_.toChar)
      .through(tokens[IO, Char])
      .map((token: Token) => {
        println(s"token: ${token}")
        token
      }).compile.toList.unsafeRunSync()
[Sample.txt](https://github.com/gnieh/fs2-data/files/12553772/Sample.txt)

I added some code to print out the chunks to see where they start and stop and found that when the chunks end with an escape character the text just before it gets duplicated on the next chunk.

    val chunkSize = 55
    val stream: fs2.Stream[IO, Byte] = fs2.io.readInputStream(IO(getClass.getResourceAsStream("/Sample.txt")), chunkSize)
    stream.map(_.toChar).chunkN(chunkSize)
      .map((chunk: Chunk[Char]) => {
        println(s"chunk: ${chunk.toList.toString()}")
        chunk
      })
      .flatMap(c => fs2.Stream.chunk(c))
      .through(tokens[IO, Char])
      .map((token: Token) => {
        println(s"token: ${token}")
        token
      }).compile.toList.unsafeRunSync()

ybasket:
Closes #515

### DIFF
--- a/json/src/main/scala/fs2/data/json/internal/JsonTokenParser.scala
+++ b/json/src/main/scala/fs2/data/json/internal/JsonTokenParser.scala
@@ -105,6 +105,7 @@ private[json] class JsonTokenParser[F[_], T, Res](
           } else if (c == '\\') {
             T.appendMarked(context, acc)
             T.advance(context)
+            T.mark(context)
             slowString_(key, StringState.SeenBackslash, 0, acc)
           } else if (c >= 0x20 && c <= 0x10ffff) {
             T.advance(context)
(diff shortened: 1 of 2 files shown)


satabin (approved):

satabin:
This is a serious bug, I will issue a bugfix release right away.

PR #526, opened 2023-09-25
Fix evaluation of stay moves

satabin:
Stay moves appear in MFT (and ESP) RHSs when a state is called on the `Self` forest. This results in the event stream processor to immediately call the state on the currently processed event. This is similar to calling a standard state on the child or next forest, but differs in the way it must evaluate the arguments.

In a standard call, the arguments must be stepped to resolve potentially nested (non self) calls. However when resolving a self call we should not apply standard calls, which should only be processed in the next step.

This is fixed in this PR.

Also this PR improves the code by:
 - using a `Vector[Expr[Out]` representation for state parameters instead of a `Map[Int, Expr[Out]]`
 - inlining more stay moves

### DIFF
--- a/finite-state/shared/src/main/scala/fs2/data/esp/ESP.scala
+++ b/finite-state/shared/src/main/scala/fs2/data/esp/ESP.scala
@@ -35,28 +35,30 @@ private[data] class ESP[F[_], Guard, InTag, OutTag](val init: Int,
                                                     val rules: DecisionTree[Guard, Tag[InTag], Rhs[OutTag]])(implicit
     F: RaiseThrowable[F]) {
 
-  def call[In, Out](env: Map[Int, Expr[Out]], q: Int, d: Int, args: List[Expr[Out]], in: Option[In])(implicit
+  def call[In, Out](env: Vector[Expr[Out]], q: Int, d: Int, args: List[Expr[Out]], in: Option[In], inline: Boolean)(
+      implicit
       In: Selectable[In, Tag[InTag]],
       Out: Conversion[OutTag, Out],
       TT: Tag2Tag[InTag, OutTag],
       G: Evaluator[Guard, Tag[InTag]]) =
     params.get(q).liftTo[Pull[F, Nothing, *]](new ESPException(s"unknown state $q")).flatMap { params =>
       if (params === args.size) {
-        args.zipWithIndex
-          .foldLeftM(env) { case (env, (arg, param)) =>
-            step(env, arg, in).map { rhs =>
-              env.updated(param, rhs)
-            }
+        args
+          .traverse { arg =>
+            if (inline)
+              Pull.pure(arg)
+            else
+              step(env, arg, in)
           }
           .flatMap { env =>
             rules
               .get(Input(q, d, in))
               .liftTo[Pull[F, Nothing, *]](new ESPException(s"no rule found for state $q at depth $d for input $in"))
-              .flatMap(eval(env, d, in, _))
+              .flatMap(eval(env.toVector, d, in, _))
           }
       } else {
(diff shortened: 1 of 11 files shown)


finite-state/shared/src/main/scala/fs2/data/mft/query/QueryCompiler.scala
```diff
@@ -58,7 +58,7 @@ private[fs2] abstract class QueryCompiler[Tag, Path] {
   /** Return the constructor tag of this pattern, or `None` if it is a wildcard. */
   def tagOf(pattern: Pattern): Option[Tag]
 
-  def compile(query: Query[Tag, Path]): MFT[NonEmptyList[Guard], Tag, Tag] = {
+  def compile(query: Query[Tag, Path], credit: Int = 50): MFT[NonEmptyList[Guard], Tag, Tag] = {
```
ybasket, line 61: I think this one could use some (scala)docs? Where does this magic value 50 come from, why would people want to pass something different…

finite-state/shared/src/main/scala/fs2/data/esp/ESP.scala
```diff
@@ -18,44 +18,47 @@ package fs2
...
   * catch all rules, no matter what the state or depth is.
   */
-private[data] class ESP[F[_], Guard, InTag, OutTag](init: Int,
+private[data] class ESP[F[_], Guard, InTag, OutTag](val init: Int,
                                                     val params: Map[Int, Int],
                                                     val rules: DecisionTree[Guard, Tag[InTag], Rhs[OutTag]])(implicit
     F: RaiseThrowable[F]) {
 
-  def call[In, Out](env: Map[Int, Expr[Out]], q: Int, d: Int, args: List[Expr[Out]], in: Option[In])(implicit
+  def call[In, Out](env: Vector[Expr[Out]], q: Int, d: Int, args: List[Expr[Out]], in: Option[In], inline: Boolean)(
```
ybasket, line 38: Any specific reason to use `Vector`? Naively, I had expected the `env` to be rather small and we have `List` for `args`, so it looks/is inconsistent.

finite-state/shared/src/main/scala/fs2/data/esp/Tag.scala
```diff
@@ -49,10 +50,36 @@ object Tag {
...
+    case Depth(d)   => show"[$d]"
+    case Name(name) => show"<$name>"
+    case Open       => "<%>"
+    case Close      => "</%>"
+    case End        => "$"
+    case Leaf       => "%"
+    case Value(v)   => show"$v"
+  }
+
+  implicit def order[T: Order]: Order[Tag[T]] = Order.from {
```
ybasket, line 70: The non-enlightened among us could use a small comment on the desired order here.

ybasket (approved):
Some smaller comments asking for clarifications, other than that, it looks good to me.

finite-state/shared/src/main/scala/fs2/data/mft/query/QueryCompiler.scala
satabin: True.

finite-state/shared/src/main/scala/fs2/data/esp/Tag.scala
satabin: It is kind of arbitrary, I just needed an order for pretty printing matched cases in ESP so that I can compare them across different versions. I will just say it is an arbitrary one for internal use.

satabin: I even made it private, as it is used only for pretty-printing for now.

Issue #413, opened 2022-11-07
Add cookbooks


satabin:
Some patterns come often when using the library, the website should have a section for this.
Among common patterns:
 - read from CSV sources to be merged and write back
 - JSON/XML formatting

It would also be nice to add a simple way to submit new ones to the website, maybe using PR templates.

PR #529, opened 2023-10-02
Add cookbook for JSON data manipulation

satabin:
This PR bootstraps the addition of examples and cookbooks in `fs2-data`. The goal is to have bigger showcases than the examples from the documentation, showing how to integrate the provided tools within a project.

This specific cookbook describes a `jq`-like CLI tool using the JSON tools from `fs2-data`.

Contributes to #413


examples/jqlike/src/main/scala/fs2/data/example/jqlike/JqLike.scala
```diff
@@ -0,0 +1,78 @@
...
+import fs2.Stream
+import fs2.data.json.jq.{Jq, JqParser}
+import fs2.data.json.{jq, render, tokens}
+import fs2.io.file.{Files, Path}
+
+object JqLike extends CommandIOApp(name = "fs2-jq", header = "A streaming implementation of a jq-like tool") {
+
+  val query: Opts[Option[String]] =
+    Opts
+      .option[String](long = "query", short = "q", help = "The query to execute on the input (default to '.')")
```
ybasket, line 32:
```suggestion
      .option[String](long = "query", short = "q", help = "The query to execute on the input (defaults to '.')")
```

examples/jqlike/src/main/scala/fs2/data/example/jqlike/JqLike.scala
```diff
@@ -0,0 +1,78 @@
...
+      .orNone
+
+  val input: Opts[Either[String, Path]] =
+    Opts
+      .option[String](long = "input", short = "i", help = "The input json string")
```
ybasket, line 37: Maybe read `stdin` instead? The only `jq` use cases I ever had involved piping JSON into it…

ybasket (approved):
Very nice! Left two small comments, but this is already good to go.

examples/jqlike/src/main/scala/fs2/data/example/jqlike/JqLike.scala
satabin: For simplicity I made the choice to have everything as explicit option and not positional arguments. So this prevents simple piping. This is an example, not a fully fledged json tool.

PR #531, opened 2023-10-05
Add a default fallback value to paths

satabin:
In some query languages and for some formats, when a path is not found, it is fine to output nothing. For instance for XML, an empty sequence of nodes is ok.

However in some cases, a default value has to be output. This is the case in the JSON query language. When the query builds an object whose value is a sub-query, then the value cannot be empty, even if the sub-query finds no element. Moreover, the query must return exactly one result. If it return zero or more than one element, then the generated JSON data is invalid.

But in the JSON query lanuage, generating no data is still valid, and happens when the path is iterated over, for instance.

This change allows for the query to parameterize for each path whether it should return a (unique) default value, or not. If not, then the previous behavior is kept, otherwise in case of no-match, then the default value is output.


ybasket (approved):
👍

PR #567, opened 2024-01-29
Add streaming pretty printing utilities

satabin:
It is inspired by the paper [_Lazy v. Yield: Incremental, Linear Pretty-printing_](https://okmij.org/ftp/continuations/PPYield/yield-pp.pdf) by Oleg Kiselyov, Simon Peyton-Jones, and Amr Sabry.

It adds indentation and alignment features to be useful in the context of streaming tree structure pretty printing (XML, JSON, ...).

Any type that has an instance of `Renderable` can be printed.

The printer guarantees that the data is printed as early as possible and the it is correct up to the first malformed tree structure.

### DIFF
--- a/json/src/main/scala/fs2/data/json/tokens.scala
+++ b/json/src/main/scala/fs2/data/json/tokens.scala
@@ -18,9 +18,16 @@ package fs2
 package data
 package json
 
+import fs2.Pure
+import fs2.data.text.render.{DocEvent, Renderable, Renderer}
+
+import scala.annotation.switch
+import scala.annotation.tailrec
+
 sealed abstract class Token(val kind: String) {
   def jsonRepr: String
 }
+
 object Token {
 
   case object StartObject extends Token("object") {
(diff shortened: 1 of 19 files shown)


json/src/main/scala/fs2/data/json/tokens.scala
```diff
@@ -55,7 +62,298 @@ object Token {
...
+    def jsonRepr: String = {
+      val rendered = new StringBuilder
+      rendered.append('"')
+      renderString(value, 0, rendered)
+      rendered.append('"')
+      rendered.result()
+    }
+  }
+
+  private final val hex = "0123456789abcdef"
```
ybasket, line 74:
```suggestion
  private final val hex = "0123456789abcdef".toArray
```

I'd expect an array of chars to be faster in index-based access (haven't benchmarked though).

json/src/main/scala/fs2/data/json/tokens.scala
```diff
@@ -55,7 +62,298 @@ object Token {
...
+  private val falseValue =
+    Stream.emit(DocEvent.Text("false"))
+
+  private final val FirstObjectKey = 0
+  private final val ObjectKey = 1
+  private final val ObjectValue = 2
+  private final val FirstArrayValue = 3
+  private final val ArrayValue = 4
+
+  implicit object renderable extends Renderable[Token] {
```
ybasket, line 133: Any specific reason to make the "pretty" renderer the implicit default and not compact? I would have expected it rather the other way round…

text/shared/src/main/scala/fs2/data/text/render/package.scala
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package fs2.data.text
+
+import fs2.Pipe
+import fs2.data.text.render.internal.StreamPrinter
+
+package object render {
+
+  def pretty[F[_], Event](width: Int = 100, indent: Int = 2)(implicit
```
ybasket, line 24: If I wanted to indent with tabs instead of spaces, I couldn't do, right? Is there maybe an easy generalisation of `indent` that allows for this?

xml/src/main/scala/fs2/data/xml/package.scala
```diff
@@ -103,15 +104,10 @@ package object xml {
       *
       * This pipe can be used when whitespace characters are not relevant to the application
       * and to make it more readable to human beings.
-      *
-      * @param collapseEmpty Whether empty tags are collapsed in a single self closing tag
-      * @param indent THe indentation string
-      * @param attributeThreshold Number of attributes above which each attribute is rendered on a new line
       */
-    def pretty[F[_]](collapseEmpty: Boolean = true,
```
ybasket, line 111: Aren't you breaking bincompat here? Looks to me like this is public, so `pretty` would need to be kept and delegate to `prettyPrint`. Also, `attributeThreshold` is gone in favour of the more general `width`, right?

ybasket (commented):
Left some questions and nitpicks, but overall, things look good 👍

json/src/main/scala/fs2/data/json/tokens.scala
satabin: True, the string random access has some overhead.

json/src/main/scala/fs2/data/json/tokens.scala
satabin: It was driven by the compact having a custom pipe, so I thought that when people use the `prettyPrint` pipe, they would actually expect pretty printing being the default.

text/shared/src/main/scala/fs2/data/text/render/package.scala
satabin: Well... How big is a tab? The pretty printing algorithm based on Wadler's (basically what I implemented here) usually tab with space and monospace fonts.

I know of other algorithms allowing for more flexible layouts (like variable width font, and maybe tabs), but TBH this did not seem worth implementing in this case.

xml/src/main/scala/fs2/data/xml/package.scala
satabin: You can't break compat if things are not released yet 😆 The pretty thing in XML is not released, I just merged it a few weeks ago.

ybasket: Fair point 😄

text/shared/src/main/scala/fs2/data/text/render/package.scala
ybasket: I briefly thought about something along the lines of
```scala
enum Indentation {
  case Spaces(amount: Int)
  case Tab(amount: Int, width: Int)
  case Custom(amount: Int, width: Int /* = value.length */, value: String)
}
```
but then there'd also be a fair argument saying it should be extendable, so an open `trait` with some methods for the relevant properties and then things escalate quickly. So yes, limiting to spaces & monospace fonts seems reasonable.

ybasket (approved):

PR #583, opened 2024-03-24
Add lichess as adopter

lenguyenthanh:


ybasket (approved):
Awesome, thank you for the PR :)

PR #604, opened 2024-06-03
Add MessagePack item parser

jarmuszz:
This PR implements a base model and a parser for the MessagePack `low` representation. There are a few tests included along with a basic benchmark. These are the results of the benchmark ran on i7-8550u:
  ```
  Benchmark                                      Mode  Cnt     Score    Error  Units
  MsgPackItemParserBenchmarks.parseMsgpackItems  avgt   10  4367.981 ± 67.376  us/op
  ```


build.sbt
```diff
@@ -462,6 +463,15 @@ lazy val cborJson = crossProject(JVMPlatform, JSPlatform, NativePlatform)
     tlVersionIntroduced := Map("3" -> "1.5.1", "2.13" -> "1.5.1", "2.12" -> "1.5.1")
   )
 
+lazy val msgpack = crossProject(JVMPlatform, JSPlatform, NativePlatform)
+  .crossType(CrossType.Full)
```
satabin, line 467: If all the sources live in the `shared` directory, let's remove this one level of indirection by using `CrossType.Pure` and move the `src` directory at the top.

msgpack/shared/src/test/scala/fs2/data/msgpack/ParserSpec.scala
```diff
@@ -0,0 +1,274 @@
...
+        A3 66 6F 6F
+        01
+        A3 62 61 7A
+        CB 3F E0 00 00 00 00 00 00
+        A9 74 69 6D 65 73 74 61 6D 70
+        D7 FF 00 11 22 33 44 55 66 77
+        """,
+       List(
+         MsgpackItem.Map(8),
+         MsgpackItem.Str(ByteVector.encodeUtf8("int").toOption.get),
```
satabin, line 233: I am a bit bothered by these `toOption.get` for some reason. We know it will not fail here, so we can avoid wrapping by using this:

```suggestion
         MsgpackItem.Str(ByteVector("int".getBytes(StandardCharsets.UTF_8))),
```

jarmuszz: Agreed, this approach is a cleaner and makes Codacy happy.

jarmuszz:
The last two commits made the parser a bit faster:
```
Benchmark                                      Mode  Cnt      Score     Error  Units
MsgPackItemParserBenchmarks.parseMsgpackItems  avgt   10  23144.104 ? 412.640  us/op
```
Which is an around 30% speedup.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/Helpers.scala
```diff
@@ -0,0 +1,87 @@
...
+      F: RaiseThrowable[F]): Pull[F, MsgpackItem, ParserResult[F, ByteVector]] = {
+    def go(count: Int, ctx: ParserContext[F], bytes: ByteVector): Pull[F, MsgpackItem, ParserResult[F, ByteVector]] = {
+      if (count <= 0) {
+        Pull.pure(ctx.toResult(bytes.reverse))
+      } else {
+        ensureChunk(ctx) { ctx =>
+          go(count - 1, ctx.next, ctx.chunk(ctx.idx) +: bytes)
+        } {
+          Pull.raiseError(new MsgpackParsingException("Unexpected end of input"))
+        }
```
satabin, line 81: The chunk might be empty (you never know) or might already have chunk byte. I think it might be even better and safer to do something like this (not compiled, but this is the idea):

```suggestion
        ensureChunk(ctx) { ctx =>
          val slice = ctx.chunk.toArraySlice
          val available = slice.size - idx
          // we accumulate either what is available, or the count
          val accumulated = Math.min(count, available)
          val newBytes = bytes ++ chunk.drop(idx).take(accumulated).toByteVector
          if (available >= count) {
            // we have enough bytes
            Pull.pure(ParserResult(slice, idx + count, rest, acc, newBytes))
          } else {
            // too short, append current bytes and continue
            go(count - available, ctx.copy(idx = slice.size), newBytes)
          }
        } {
          Pull.raiseError(new MsgpackParsingException("Unexpected end of input"))
        }
```

satabin: Also, `ByteVector` has a quite efficient append (`++`) so you do not have to prepend bytes one by one and reverse, this will even probably be less efficient.

satabin: I use `toArraySlice` here because it has O(1) `drop` and `take`.

jarmuszz: Great catch! This should also make the function a *lot* faster. I will implement this later today.

jarmuszz:
I am considering adding explicit 0-length checks into format parsers and making the `count <= 0` in `requireBytes` raise instead of emiting an empty `ByteVector`. This way we could differentiate between unwanted and wanted behavior (e.g. 0-length string is valid whereas integer is not).

msgpack/src/main/scala/fs2/data/msgpack/low/internal/Helpers.scala
satabin: Did that bring any speedup in the end?

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemParser.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  /** Parses a single item from the stream.
+    */
+  private def parseItem[F[_]](ctx: ParserContext[F])(implicit
+      F: RaiseThrowable[F]): Pull[F, MsgpackItem, ParserContext[F]] = {
+    requireOneByte(ctx) flatMap { res => {
+      val byte = res.result
+      val ctx = res.toContext
+
+      byte match {
```
satabin, line 37: A nice way to have for sure something efficient, is to annotate it with `@switch`. This ensures that the emitted code is a switch. If it cannot be compiled as such it will emit a warning. For this to work, you will need to use `Int` instead of `Byte`. You can get it efficiently by doing:
```suggestion
      ((byte & 0xff): @switch) match {
```

And by removing all the `toByte` in the `Headers` object.

satabin:
Note: the workflow are not up-to-date with the build, be sure to run `sbt prePR` to check that all static checks are passing.

build.sbt
```diff
@@ -462,6 +463,15 @@ lazy val cborJson = crossProject(JVMPlatform, JSPlatform, NativePlatform)
...
   )
 
+lazy val msgpack = crossProject(JVMPlatform, JSPlatform, NativePlatform)
+  .crossType(CrossType.Pure)
+  .in(file("msgpack"))
+  .settings(commonSettings)
+  .settings(
+    name := "fs2-data-msgpack",
+    description := "Streaming MessagePack library"
+  )
```
ybasket, line 473: I think we'll need the `jsSettings` the other modules have here as well.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/FormatParsers.scala
```diff
@@ -0,0 +1,131 @@
...
+          seconds = res.result
+        } yield res.toContext.prepend(MsgpackItem.Timestamp96(nanosec, seconds))
+      case _ => Pull.raiseError(new MsgpackParsingException(s"Invalid timestamp length: ${length}"))
+    }
+  }
+
+  def parseFixExt[F[_]](length: Int, ctx: ParserContext[F])(implicit
+      F: RaiseThrowable[F]): Pull[F, MsgpackItem, ParserContext[F]] = {
+    requireOneByte(ctx) flatMap {
+      res => {
```
ybasket, line 83:
```suggestion
    requireOneByte(ctx).flatMap { res => 
```

No need for an extra block after `res =>`. Also, I'd prefer if we could avoid infix `flapMap` (and alike) for a) consistency and b) because Scala 3 only wants infix usage of methods defined explicitly as infix. Not keen on having to fix all occurrences with an upcoming Scala update…this obviously applies to all occurrences, just picked this line as an example.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemParser.scala
```diff
@@ -0,0 +1,114 @@
...
+        case Headers.Str8      => parseString(1, ctx)
+        case Headers.Str16     => parseString(2, ctx)
+        case Headers.Str32     => parseString(4, ctx)
+        case Headers.Array16   => parseArray(2, ctx)
+        case Headers.Array32   => parseArray(4, ctx)
+        case Headers.Map16     => parseMap(2, ctx)
+        case Headers.Map32     => parseMap(4, ctx)
+        case _ => {
+          // Positive fixint
+          if ((byte & 0x80) == 0) {
```
ybasket, line 71: If it doesn't interfere with the suggestion about `@switch` above, I think this would be more readable if the if-else construct was inlined in to the pattern matching, avoiding the nesting.

msgpack/src/main/scala/fs2/data/msgpack/package.scala
ybasket, line 1: If it's empty, what's the reason to have this in the first place?

ybasket (commented):
Looks very promising, just some details to iron out :)

msgpack/src/main/scala/fs2/data/msgpack/low/internal/Helpers.scala
jarmuszz: Yes, quite a big one actually!
```
Benchmark                                      Mode  Cnt     Score    Error  Units
MsgPackItemParserBenchmarks.parseMsgpackItems  avgt   10  4659.946 ? 66.764  us/op
```
It seems that the previous `requireBytes` implementation has been a major bottleneck.

msgpack/src/main/scala/fs2/data/msgpack/package.scala
jarmuszz: Thanks. I left it there by a mistake :).

msgpack/src/main/scala/fs2/data/msgpack/low/internal/Helpers.scala
satabin: Yes they were the all lot of instantiation I spotted at the beginning.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemParser.scala
jarmuszz: Done in 3272920. I have actually tried to do this before but I missed the fact that `@switch` requires expression to be of type `Int` :smile:.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/FormatParsers.scala
jarmuszz: Makes sense! Fixed in afa91e1.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemParser.scala
```diff
@@ -0,0 +1,109 @@
...
+        case Headers.NeverUsed => Pull.raiseError(new MsgpackParsingException("Reserved value 0xc1 used"))
+        case Headers.False     => Pull.pure(ctx.prepend(MsgpackItem.False))
+        case Headers.True      => Pull.pure(ctx.prepend(MsgpackItem.True))
+        case Headers.Bin8      => parseBin(1, ctx)
+        case Headers.Bin16     => parseBin(2, ctx)
+        case Headers.Bin32     => parseBin(4, ctx)
+        case Headers.Ext8      => parsePlainExt(1, ctx)
+        case Headers.Ext16     => parsePlainExt(2, ctx)
+        case Headers.Ext32     => parsePlainExt(4, ctx)
+        case Headers.Float32   => parseSimpleType(MsgpackItem.Float32.apply)(4, ctx)
```
satabin, line 49: It's a minor question of style, but I tend to prefer this syntax:

```suggestion
        case Headers.Float32   => parseSimpleType(MsgpackItem.Float32(_))(4, ctx)
```

satabin:
This looks good to me for a first PR. Can you provide the latest benchmarks in the description?

jarmuszz:
> Regarding the compilation failure, this is due to the fact that pattern matches with guards cannot be compiled to a switch. So it emits a warning, which is an error i the CI.
> 
> To keep the efficient switch compilation, you'll need to put back the if structure within the catch-all case in the end.

Interestingly, this warning has not been showing up in my local builds. After some digging, I found out that the match compiles just fine on scala 2.13 but fails on 3.3.3. Not sure if this is an intended behavior or a bug.

satabin:
> > Regarding the compilation failure, this is due to the fact that pattern matches with guards cannot be compiled to a switch. So it emits a warning, which is an error i the CI.
> > To keep the efficient switch compilation, you'll need to put back the if structure within the catch-all case in the end.
> 
> Interestingly, this warning has not been showing up in my local builds. After some digging, I found out that the match compiles just fine on scala 2.13 but fails on 3.3.3. Not sure if this is an intended behavior or a bug.

Behaviors are different on that matter in scala 3, and generally speaking scala 3 is more sound. So I tend to use it as the biggest common denominator.

build.sbt
```diff
@@ -31,7 +31,7 @@ val diffsonVersion = "4.6.0"
 val literallyVersion = "1.1.0"
 val weaverVersion = "0.8.4"
 
-ThisBuild / tlBaseVersion := "1.11"
+ThisBuild / tlBaseVersion := "1.12"
```
satabin, line 34: This should not be changed. This is only changed after a release. The base version is the latest published one on which your changes are based. And this still is the 1.11.x series.

jarmuszz: My bad. Thanks for the catch :slightly_smiling_face:

satabin: No worries, the build is a bit arcane at times.

satabin (approved):
Congrats on the first fully green build! It looks good for me.

satabin:
@ybasket Are you fine with the PR?

ybasket (approved):

PR #612, opened 2024-06-23
Handle case class defaults better

### LINKED ISSUE #607: Support for / respect of default arguments in case class decoders?
Hello,

I'm new to `fs2-data` and am confused by the last section of the [documentation for the `generic` module](https://fs2-data.gnieh.org/documentation/csv/generic/). Specifically, it's mentioned that...

> Both automatic and semi-automatic decoders support also default values when decoding

...but then it looks in fact like if a row's missing a value in a slot that _could_ be populated by a default value when building an instance of the target case class, that value's not used and instead a `DecoderError` arises. Here's the snippet:
```scala
import fs2.data.csv.generic.auto._

case class MyRowDefault(i: Int = 42, j: Int, s: String)

val decoded = stream.through(decodeUsingHeaders[MyRowDefault]())
// decoded: Stream[[x]Fallible[x], MyRowDefault] = Stream(..)
decoded.compile.toList
// res11: Either[Throwable, List[MyRowDefault]] = Left(
//   value = fs2.data.csv.DecoderError: unable to decode '' as an integer in line 3
// )
```

That doesn't feel like "support" for default values, but rather something like "tolerance" of their existence and incompatibility with an encounter of a row missing a value in a slot for which there's a default. Is the example actually the intended behavior?

ybasket:
Always parse empty cells to their default values if one is defined, allowing to use defaults with types that can't be parsed from an empty cell, like numbers. Comes at the expense of not being able to parse empty cells as-is any longer if a default exist.

Closes #607


csv/generic/shared/src/test/scala-2/fs2/data/csv/generic/CsvRowDecoderDefaultsTest.scala
```diff
@@ -0,0 +1,34 @@
...
+  pureTest("case classes should be handled properly with default value and empty cell") {
+    expect(testDecoder(csvRowDefaultI) == Right(Test(0, "test", Some(42))))
+  }
+
+  pureTest("case classes should be handled properly with default value and missing column") {
+    expect(testDecoder(csvRowNoI) == Right(Test(0, "test", Some(42))))
+  }
+
+  pureTest("case classes should be handled properly with default string value and empty cell") {
+    expect(testDefaultsDecoder(csvRowEmptyJ) == Right(TestDefaults(1, "foo")))
```
ybasket, line 31: Note that the default `"foo"` is expected even though a `String` cell could handle the empty cell. While I think it's the better behaviour for users as they now can handle `Int` or any other type that can't be parsed from empty input via default, it is a change that we should highlight in the release notes. Users that need to handle empty cells may no longer define a default (except for `""` itself) for those in their `case class`.

csv/generic/shared/src/main/scala-2/fs2/data/csv/generic/internal/MapShapedCsvRowDecoder.scala
```diff
@@ -74,6 +74,7 @@ private[generic] trait LowPriorityMapShapedCsvRowDecoder1 {
                           default: Option[Head] :: DefaultTail,
                           anno: Anno :: AnnoTail): DecoderResult[FieldType[Key, Head] :: Tail] = {
         val head = row(anno.head.fold(witness.value.name)(_.name)) match {
+          case Some(head) if head.isEmpty && default.head.nonEmpty => default.head.get.asRight
```
satabin, line 77: I'd love to avoid the `.get` if possible. Something like:

```suggestion
          case Some(head) if head.isEmpty && default.head.nonEmpty => default.head.toRight(new DecoderError("Should not happen", row.line))
```

csv/generic/shared/src/test/scala-2/fs2/data/csv/generic/CsvRowDecoderDefaultsTest.scala
satabin: Thinking about it, I would add a clear note in the documentation for this caveat.

ybasket: Done in ad960a3f20d869f234dd349b4cc016077322a6cc

satabin (approved):
LGTM, you juste need to run `sbt prePR` and you're good to go!

Issue #603, opened 2024-06-03
Add MessagePack support


jarmuszz:
This ticket will be the root of the [MessagePack](https://msgpack.org/) implementation process.

Similarly to `fs2-data-cbor`, the `fs2-data-msgpack` should be split into two modules:
- The `low` module which will work on a flat model that closely mimics the raw MessagePack format. Data in this model will not be converted into language-level values but rather kept in it's byte form.
- The `high` module that will work on an AST typed accordingly to the [implementation specification](https://github.com/msgpack/msgpack/blob/master/spec.md#serialization-type-to-format-conversion). The parsing and serialization in this module should most likely be built on top of the `low` module.

We can split the process of implementing the format into the following parts:
- [ ] `low`
  - [x] model and parser (#604)
  - [X] msgpack encoder (#628)
  - [X] validator (#628)
  - [ ] documentation
- [ ] `high`
  - [ ] model and parser
  - [ ] msgpack encoder
  - [ ] documentation

PR #618, opened 2024-07-07
Improve MessagePack model

jarmuszz:
Current implementation of the MessagePack model allows for creation of malformed objects as the `ByteVector` class (which underlays most of the types in the model) has a maximum length of `Long.MaxValue` bytes which is bigger than the [specification](https://github.com/msgpack/msgpack/blob/master/spec.md) defines for some types. 

This PR proposes following changes:
- `Float32(ByteVector)` -> `Float32(Float)`
- `Float64(ByteVector)` -> `Float64(Double)`
- `Timestamp32(ByteVector)` -> `Timestamp64(Int)`
- `Timestamp64(ByteVector, ByteVector)` -> `Timestamp64(Long)` with `nanoseconds` and `seconds` fields
- `Timestamp96(ByteVector, ByteVector)` -> `Timestamp96(Int, Long)` 

Ref. #603


jarmuszz:
I have currently held back from changing the integer types. `UnsignedInt` would have to hold a `BigInt` in order to keep a sign which is not quite performant. We could also bit cast it into a `Long` but I think it would be confusing for users.

msgpack/src/main/scala/fs2/data/msgpack/low/model.scala
```diff
@@ -40,9 +40,18 @@ object MsgpackItem {
...
-  case class Timestamp96(nanoseconds: ByteVector, seconds: ByteVector) extends MsgpackItem
+  case class Timestamp32(seconds: Int) extends MsgpackItem
+
+  /** Stores data in a 30-bit [[nanoseconds]] and a 34-bit [[seconds]] fields, both of which are accessible as class
+    * attributes. To ensure valid data length at the type level, both fields are constructed from a single 64-bit
+    * [[combined]] variable.
+    * @param combined [[nanoseconds]] and [[seconds]] combined into a signle 64-bit value
+    */
+  case class Timestamp64(combined: Long) extends MsgpackItem {
+    val nanoseconds: Int = (combined >> 34).toInt // we are sure that (x: Long) >> 34 fits in an int
```
satabin, line 51: When using the arithmetic right shift, it is better to apply a mask before converting.

```suggestion
    val nanoseconds: Int = (0xffffffffL & (combined >> 34)).toInt // we are sure that (x: Long) >> 34 fits in an int
```

satabin: (note that this mask has only 7 f's and not 8 like the other one)

jarmuszz: I guess without the mask we could end up padding with 1s if the number is negative? Is that right?

satabin: Yes, this is what can happen.

satabin (approved):
Looks good to me, let's see if @ybasket has some comments.

ybasket:
> Looks good to me, let's see if @ybasket has some comments.

No, was only waiting for the small fix you had requested. LGTM

PR #622, opened 2024-07-27
Emit an error if row and header size mismatch

### LINKED ISSUE #621: `encodeGivenHeaders` allows different number of headers than the number of columns with actual data
when using `encodeGivenHeaders[A]`, if the number of elements in the `headers` argument is different to the number of columns that are encoded for values of  `A`, it is probably something the user wants to know about and the Stream should probably fail with an error instead of encoding an invalid csv

satabin:
Fixes #621 

A new method is created and the old one is deprecated to keep binary compatibility.

### DIFF
--- a/csv/shared/src/main/scala/fs2/data/csv/package.scala
+++ b/csv/shared/src/main/scala/fs2/data/csv/package.scala
@@ -198,6 +198,10 @@ package object csv {
   }
 
   /** Encode a specified type into a CSV prepending the given headers. */
+  @deprecated(
+    message =
+      "Emits incorrect data if rows have a different length than headers. Please use `encodeWithGivenHeaders` instead.",
+    since = "fs2-data 1.11.1")
   def encodeGivenHeaders[T]: PartiallyAppliedEncodeGivenHeaders[T] =
     new PartiallyAppliedEncodeGivenHeaders[T](dummy = true)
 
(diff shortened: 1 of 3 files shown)


csv/shared/src/main/scala/fs2/data/csv/package.scala
```diff
@@ -217,6 +221,27 @@ package object csv {
     }
   }
 
+  /** Encode a specified type into a CSV prepending the given headers. */
+  def encodeWithGivenHeaders[T]: PartiallyAppliedEncodeWithGivenHeaders[T] =
```
ybasket, line 225: Wdyt about renaming to ``encodeWithHeaders`? We already have `encodeWithoutHeaders`, so the naming feels a bit more consistent then the `WithGiven`.

No strong opinion on this though, just noticed it, decide as you see fit.

csv/shared/src/test/scala/fs2/data/csv/RowGeneratorTest.scala
```diff
@@ -0,0 +1,51 @@
...
+object RowGeneratorTest extends SimpleIOSuite {
+
+  pureTest("Emit error on wrong row size (#621)") {
+    val input = List(
+      Row(NonEmptyList.of("a", "b", "c"), Some(1)),
+      Row(NonEmptyList.of("d", "e"), Some(2)),
+      Row(NonEmptyList.of("f", "g", "h", "i"), Some(3)),
+      Row(NonEmptyList.of("j", "k", "l"), Some(4))
+    )
+    val headers = NonEmptyList.of("first", "second", "third")
```
ybasket, line 31:
```suggestion
    val headers = NonEmptyList.of("first", "second", "third")
    val input = List(
      Row(NonEmptyList.of("a", "b", "c"), Some(1)),
      Row(NonEmptyList.of("d", "e"), Some(2)),
      Row(NonEmptyList.of("f", "g", "h", "i"), Some(3)),
      Row(NonEmptyList.of("j", "k", "l"), Some(4))
    )
```

I find this test case slightly annoying to read, but except for this reordering (so it matches output order) and maybe adding some `unapplySeq` in cats itself, I found no better.

ybasket (approved):

csv/shared/src/test/scala/fs2/data/csv/RowGeneratorTest.scala
ybasket: One more thought on this snippet (not to be done in this PR though): Maybe a `row"…, …, …"` string interpolator could be nice, especially for our tests?

csv/shared/src/main/scala/fs2/data/csv/package.scala
satabin: I wanted to keep consistent naming between the high and low level naming. Since the low level naming cannot be `writeWithHeaders`(this is the deprecated one) and high-level cannot be `encodeGivenHeaders` (this is the deprecated one), I settled for merging both.

csv/shared/src/test/scala/fs2/data/csv/RowGeneratorTest.scala
ybasket: Tried adding `unapplySeq` to cats, unfortunately it doesn't work because `NonEmptyList` is a `case class` and hence has a generated `unapply` that takes precedence. Would need a dedicated extractor object.

PR #628, opened 2024-08-18
Add MessagePack item serializer and validator

jarmuszz:
This pull request brings serialization and validation to the `msgpack.low` module.

Serialization is provided for the outside use through `tobinary: Pipe[F, MsgpackItem, Byte]` and `toNonValidatedBinary: Pipe[F, MsgpackItem, Byte]` functions. Akin to how the `cbor.low` serialization is constructed, serialization without validation can potentialy produce malformed data.

Validation API is also exposed through `validated: Pipe[F, MsgpackItem, MsgpackItem]`.


jarmuszz:
The initial concept of having two serializers was dropped as they ran at a nearly identical time :smiley:.

These are the current benchmarks on i7-8550U:
```
Benchmark                                       Mode  Cnt     Score    Error  Units
MsgPackItemSerializerBenchmarks.serialize       avgt   10  3337.518 ± 39.333  us/op
MsgPackItemSerializerBenchmarks.withValidation  avgt   10  5107.575 ± 65.148  us/op
```

Validation seems to slow things down a bit and I think that maybe it is possible to make it a little faster.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemSerializer.scala
```diff
@@ -0,0 +1,247 @@
...
+      else if (bs.size <= 8)
+        o.push(ByteVector(Headers.Uint64) ++ bs.padLeft(8))
+      else
+        Pull.raiseError(new MalformedUintError)
+
+    case MsgpackItem.SignedInt(bytes) =>
+      val bs = bytes.dropWhile(_ == 0)
+      if (bs.size <= 1)
+        // positive fixint or negative fixint
+        if ((bs & hex"7f") == bs || (bs & hex"c0") == hex"c0")
```
satabin, line 103: We could extract the constants (e.g. `hex"7f"`) into object fields to avoid reallocating a `ByteVector` each time.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemSerializer.scala
```diff
@@ -0,0 +1,247 @@
...
+  private case class SerializationContext[F[_]](out: Out[F],
+                                                chunk: Chunk[MsgpackItem],
+                                                idx: Int,
+                                                rest: Stream[F, MsgpackItem])
+
+  /** Buffers [[Chunk]] into 4KiB segments before calling [[Pull.output]].
+    *
+    * @param contents buffered [[Chunk]]
+    */
+  private class Out[F[_]](contents: Chunk[Byte]) {
```
satabin, line 46: I like the approach, however, this creates a new `Out` instance for each incoming item. If you run the benchmark with allocation and CPU flame graphs, does it appear to be a bottleneck?

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemValidator.scala
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  type ValidationContext = (Chunk[MsgpackItem], Int, Long, List[Expect])
+
+  def pipe[F[_]](implicit F: RaiseThrowable[F]): Pipe[F, MsgpackItem, MsgpackItem] = { in =>
+    def step1(chunk: Chunk[MsgpackItem], idx: Int, position: Long): Pull[F, MsgpackItem, Option[Expect]] =
+      chunk(idx) match {
+        case MsgpackItem.UnsignedInt(bytes) =>
+          if (bytes.size > 8) Pull.raiseError(new ValidationErrorAt(position, "Unsigned int exceeds 64 bits"))
+          else Pull.pure(None)
```
satabin, line 39: Since we expect to have many of them, I would create a constant like `private final val PullNone = Pull.pure(None)`. This would avoid a lot of allocations.

msgpack/src/main/scala/fs2/data/msgpack/low/package.scala
```diff
@@ -18,11 +18,22 @@ package fs2
...
-import low.internal.ItemParser
+import low.internal.{ItemParser, ItemSerializer, ItemValidator}
 
 /** A low-level representation of the MessagePack format.
   */
 package object low {
   def items[F[_]](implicit F: RaiseThrowable[F]): Pipe[F, Byte, MsgpackItem] =
     ItemParser.pipe[F]
+
+  /** Alias for `bytes(validated = true)`
```
satabin, line 29: Where is `bytes` defined?

satabin:
Thanks a ton for this new contribution. I had a first look at it and left a few comments. But it looks great already. :clap:

msgpack/src/main/scala/fs2/data/msgpack/low/package.scala
jarmuszz: Forgot to change this in [309569e](https://github.com/gnieh/fs2-data/pull/628/commits/309569e48961642af378989709a29637f377759f) :smile:

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemValidator.scala
jarmuszz: What is the difference between having `final val` and `val` in an object/method? I know that this keyword matters when defining a class field but I'm not aware of it's uses in other contextxs

benchmarks/src/main/scala/fs2/data/benchmarks/MsgPackItemSerializerBenchmarks.scala
```diff
@@ -0,0 +1,71 @@
...
+@Warmup(iterations = 3, time = 2)
+@Measurement(iterations = 10, time = 2)
+class MsgPackItemSerializerBenchmarks {
+  val msgpackItems: List[fs2.data.msgpack.low.MsgpackItem] = {
+    val bytes =
+      fs2.io
+        .readClassLoaderResource[SyncIO]("twitter_msgpack.txt", 4096)
+        .through(fs2.text.utf8.decode)
+        .compile
+        .string
```
ybasket, line 41: If I'm not mistaken, this destroys all chunking, you'll work with one gigantic `Chunk` – not what real life code should usually do. I think you can simply combine those two streams roughly along the lines of

```scala
fs2.io
        .readClassLoaderResource[SyncIO]("twitter_msgpack.txt", 4096)
        .through(fs2.text.utf8.decode)
        .map(str => Chunk.byteVector(ByteVector.fromHex(str).get))
        .unchunks
        .through(fs2.data.msgpack.low.items[SyncIO])
        .compile
        .toList
        .unsafeRunSync()
```

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemSerializer.scala
```diff
@@ -0,0 +1,247 @@
...
+package internal
+
+import scodec.bits._
+
+private[low] object ItemSerializer {
+  class MalformedItemError extends Error("item exceeds the maximum size of it's format")
+  class MalformedStringError extends MalformedItemError
+  class MalformedBinError extends MalformedItemError
+  class MalformedIntError extends MalformedItemError
+  class MalformedUintError extends MalformedItemError
```
ybasket, line 30: Are those mapped somewhere? Because as they're within a `private[low]`, users won't have access to them and can't distinguish what failed. Maybe something we'd want to allow? The CSV module has public exception types.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemSerializer.scala
```diff
@@ -0,0 +1,247 @@
...
+    case MsgpackItem.Str(bytes) =>
+      if (bytes.size <= 31) {
+        o.push(ByteVector.fromByte((0xa0 | bytes.size).toByte) ++ bytes)
+      } else if (bytes.size <= Math.pow(2, 8) - 1) {
+        val size = ByteVector.fromByte(bytes.size.toByte)
+        o.push(ByteVector(Headers.Str8) ++ size ++ bytes)
+      } else if (bytes.size <= Math.pow(2, 16) - 1) {
+        val size = ByteVector.fromShort(bytes.size.toShort)
+        o.push(ByteVector(Headers.Str16) ++ size ++ bytes)
+      } else if (fitsIn(bytes.size.toInt, 32)) {
```
ybasket, line 131: Isn't this guaranteed to be `true`? `.toInt` will truncate the size to `2 ^31 - 1` which always fits in 32 bits, doesn't it? 🤔 

Also, while it's cool to have a generic function here, just comparing the size to `Int.MaxValue` might be much more efficient.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemSerializer.scala
```diff
@@ -0,0 +1,247 @@
...
+  class MalformedItemError extends Error("item exceeds the maximum size of it's format")
+  class MalformedStringError extends MalformedItemError
+  class MalformedBinError extends MalformedItemError
+  class MalformedIntError extends MalformedItemError
+  class MalformedUintError extends MalformedItemError
+
+  /** Checks whether integer `x` fits in `n` bytes. */
+  @inline
+  private def fitsIn(x: Int, n: Long): Boolean =
+    java.lang.Integer.compareUnsigned(x, (Math.pow(2, n.toDouble).toLong - 1).toInt) <= 0
```
ybasket, line 35: If this is really necessary, see comment below, then I believe we can improve its performance by using a bit shift. Taking both arguments as `Long`, that'd be 
```scala
java.lang.Long.compareUnsigned(x, (1L << n) - 1).toInt) <= 0
```

(needs some testing though, I'm no bit shift pro and probably got an off-by-one somewhere)

msgpack/src/test/scala/fs2/data/msgpack/ValidationSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+    Stream
+      .emits(cases)
+      .evalMap { case (lhs, rhs) =>
+        Stream
+          .emits(lhs)
+          .through(low.toBinary[F])
+          .compile
+          .drain
+          .map(_ => failure(s"Expected error for item ${lhs}"))
+          .handleError(expect.same(_, rhs))
```
ybasket, line 56: Minor: Understanding this bit is a bit tricky as it's unusual to expect the exception. I think `redeem` could make things more clear:
```suggestion
          .redeem(expect.same(_, rhs), _ => failure(s"Expected error for item ${lhs}"))
```

ybasket (commented):
Looks good overall 👏 , left only some smaller comments.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemValidator.scala
satabin: In this case it does not matter, generated bytecode will be the same. It's just some sort of muscle memory from my java days that came back to life :laughing:

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemSerializer.scala
jarmuszz: I have decided that having separate exception for each type is a bit too granular. Instead, there's a general `MsgpackMalformedItemException` class since [482bf9e](https://github.com/gnieh/fs2-data/pull/628/commits/482bf9e5fbbe4bf9cf244d1fa2fc0d6b081ed598).

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemSerializer.scala
jarmuszz: I completely got rid of the `fitsIn` function as we don't use `Int`s in this context anymore.

satabin:
Hi @jarmuszz. I am not sure anymore what the state of this PR is. Is it ready to be re-reviewed or do you still want to address some comments or improvements?

jarmuszz:
Hi @satabin,
Sorry for the confusion and a late reply.

I'll try to address these comments this weekend. I've been trying to combine work with full-time studies which turned out to be more time sinking than I imagined :smiley:. I hope that in the following weeks I'll be working on the fs2-data-msgpack project in a more regular fashion since I'm resigning from the job.

msgpack/src/test/scala/fs2/data/msgpack/ValidationSpec.scala
jarmuszz: Done in fce1083.

benchmarks/src/main/scala/fs2/data/benchmarks/MsgPackItemSerializerBenchmarks.scala
jarmuszz: Because the `twitter_msgpack.txt` file contained space separated hex representations of bytes, the `fs2.text.utf8.decode` pipe could slice some bytes in half between two chunks. E.g. `"A1 B2 C3"` would be converted into `Chunk(0xa1, 0x0b)` and `Chunk(0x02, 0xc3)` if our byte stream had a chunk size equal to 4.

We could remove spaces from input data but even then we would rely on the chunk size begin a multiple of two which I think is a bit fragile and hard to debug.

Instead, I converted the test data into a raw binary form which is loaded into memory as a chunked stream before benchmarks are run. The implementation looks a bit clunky but I'm not aware of any better way to achieve this.

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemSerializer.scala
jarmuszz: It doesn't seem like a significant bottleneck to me.

Here's a screenshot from VisualVM memory profiler:
image

I cannot find any reference to `Out`'s methods in CPU samples so either I'm doing something wrong or these methods get optimized out :)

satabin: I was thinking about runnin the benchmark with async-profiler outputting flamegraph. Something like:

```shell
$ sbt
> project benchmarksJVM
> Jmh/run fs2.data.benchmarks.MsgPackItemParserBenchmarks -prof "async:libPath=/path/to/libasyncProfiler.so;output=flamegraph;event=cpu;alluser=true"
```

jarmuszz: It takes around 10% of CPU and 17% of allocation graphs. 

Making the`Out` class an implicit around `Chunk[Byte]` lowers the results into 7% and 15% respectively. I'm not sure how could we optimize this more.

image
image

satabin: Time and memory seem to be dominated by `ByteVector` creation at every step. I need to review again these creations, and why they occur. But I could live with merging the current state and improving later on.

satabin:
Hi @jarmuszz. Sorry I am replying late again. What is the state of this PR? Do you still plan on working on some parts? Or is it ready to be merged when approved?

jarmuszz:
Hi, I think it is ready to be merged if approved.

satabin (approved):
Reviewed it again, and it looks good to me as a first shot. Thanks again @jarmuszz!

msgpack/src/test/scala/fs2/data/msgpack/ValidationSpec.scala
```diff
@@ -0,0 +1,112 @@
...
+      .evalMap { case (lhs, rhs) =>
+        Stream
+          .emit(lhs)
+          .through(low.toBinary[F])
+          .compile
+          .drain
+          .redeem(expect.same(_, rhs), _ => failure(s"Expected error for item ${lhs}"))
+      }
+      .compile
+      .foldMonoid
```
ybasket, line 40: We can just delegate to `validation` here:
```suggestion
    validation[F](cases.map { case (i, t) => List(i) -> t }*)
```

msgpack/src/test/scala/fs2/data/msgpack/SerializerSpec.scala
```diff
@@ -0,0 +1,190 @@
...
+
+    val out = for {
+      data <- Stream.emits(cases)
+      pre <- round(data)
+      processed <- round(pre)
+    } yield {
+      if (processed == pre)
+        success
+      else
+        failure(s"Serializer should be fixpoint for ${pre} but it emitted ${processed}")
```
ybasket, line 184:
```suggestion
      expect(processed == pre, s"Serializer should be fixpoint for ${pre} but it emitted ${processed}")
```

msgpack/src/main/scala/fs2/data/msgpack/low/internal/ItemSerializer.scala
ybasket: I think we can make this a `final class(…) extends AnyVal` (moving the `val limit` to its companion), that could save us some allocations.

ybasket (approved):

Issue #634, opened 2024-09-22
Significant performance degradation with new json rendering Introduced in 1.11


recons:
In our case, we are dealing with a large stream of big objects, which are serialized into json and written to a file. After upgrading to 1.11, we noticed higher execution time, an increase in cpu usage and more time spent in gc.
 
 Below is a minimal code snippet to demonstrate the issue:
 ```scala
fs2.Stream.range(0, 10000000)
      .map(Json.fromInt)
      .covary[IO]
      .through(ast.tokenize)
      .through(wrap.asTopLevelArray)
      .through(render.compact)
      .compile
      .drain
      .timed
      .map(_._1.toMillis.toString())
      .flatMap(IO.println)
 ```

On my machine (MBP 16 2019, 2.6 GHz 6-Core Intel Core i7), the above code takes:

- **1.11**: 81257 ms
- **1.10**: 20909 ms

While this isn't a full-fledged analysis, we suspect that the regression is related to the new rendering in 1.11. We plan to stick with version 1.10 for now, but further insight would be appreciated.
Perhaps I'm missing something? Could this performance hit be expected under certain conditions, or is it likely a regression?

satabin:
Hi. We changed the rendering in 1.11, which is probably the cause. I am thinking of reverting the compact rendering to what it was in 1.10. The introduced change is too impatfull for this use case.

recons:
Hi, thanks for your reply.

Unfortunately, it's not just the compact rendering that's been affected. In the same use case, when switching to render.prettyPrint, the process doesn't even complete after 5 minutes. It also produces numerous warnings, such as:
```
[WARNING] Your app's responsiveness to a new asynchronous event (such as a new connection, an upstream response, or a timer) was in excess of 100 milliseconds. Your CPU is probably starving. Consider increasing the granularity of your delays or adding more cedes. This may also be a sign that you are unintentionally running blocking I/O operations (such as File or InetAddress) without the blocking combinator.
```
On the other hand, using the deprecated render.pretty takes only 23 seconds and completes without any warnings at all.

satabin:
The pretty printer has been entirely changed as well. It looks like some ceding is necessary in its implementation. I will investigate.

satabin:
Hi @recons,

I ran some benchmarks and identified several sources for slow downs. I am now working on fixing them. I will open a PR as soon as I have something usable, so that you can try it with your use case.

satabin:
Fixed by #641.

PR #641, opened 2024-10-24
Improve stream rendering performances

satabin:
This change addresses #634 in two ways:
 - A change to the generic pretty printer that allows to avoid:
   - Boxing of integers when computing the layout
   - Instantiating many tuples that are immediately discarded
   - Creating intermediate chunks by fusing the annotation and rendering phases
 - Re-introduce a direct compact rendering for JSON, that is way simpler than leveraging the generic printer with no groups

By doing so, the compact rendering should be back to the performance in 1.10 and the pretty case is improved a bit.

Here are some benchmark results with the compact rendering in 1.11.1 as the baseline. 1.11.2 represents the results with this change.

```mermaid
xychart-beta
  title "Rendering of int array"
  x-axis ["Pretty 1.10", "Pretty 1.11.1", "Pretty 1.11.2", "Compact 1.11.1", "Compact 1.11.2"]
  y-axis "Factor" 0 --> 1.2
  bar [0.02, 1.12, 0.71, 1, 0.004]
```

```mermaid
xychart-beta
  title "Rendering of int object"
  x-axis ["Pretty 1.10", "Pretty 1.11.1", "Pretty 1.11.2", "Compact 1.11.1", "Compact 1.11.2"]
  y-axis "Factor" 0 --> 1.2
  bar [0.01, 1.11, 0.85, 1, 0.01]
```

### DIFF
--- a/benchmarks/src/main/scala/fs2/data/benchmarks/PrinterBenchmarks.scala
+++ b/benchmarks/src/main/scala/fs2/data/benchmarks/PrinterBenchmarks.scala
@@ -1,0 +1,85 @@
+package fs2.data.benchmarks
+
+import cats.effect.SyncIO
+import cats.effect.IO
+import fs2.data.json.Token
+import fs2.data.json.circe.*
+import fs2.{Fallible, Stream}
+import io.circe.Json
+import org.openjdk.jmh.annotations.*
+import org.openjdk.jmh.infra.Blackhole
+import cats.effect.unsafe.implicits.global
+
+import java.util.concurrent.TimeUnit
+
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@BenchmarkMode(Array(Mode.AverageTime))
+@State(org.openjdk.jmh.annotations.Scope.Benchmark)
+@Fork(value = 1)
+@Warmup(iterations = 15, time = 5)
+@Measurement(iterations = 10, time = 5)
+class PrinterBenchmarks {
+
+  val intArrayStream =
+    Stream.emits(
+      Token.StartArray ::
+        (List
+          .range(0, 1000000)
+          .map(i => Token.NumberValue(i.toString())) :+ Token.EndArray))
+
+  val objectStream =
+    Stream.emits(
+      Token.StartObject ::
+        (List
+          .range(0, 1000000)
+          .flatMap(i => List(Token.Key(s"key:$i"), Token.NumberValue(i.toString()))) :+ Token.EndObject))
+
+  @Benchmark
+  def intArrayCompact(bh: Blackhole) =
+    bh.consume(
+      intArrayStream
+        .through(fs2.data.json.render.compact)
+        .compile
+        .drain)
+
+  @Benchmark
+  def objectCompact(bh: Blackhole) =
+    bh.consume(
+      objectStream
+        .through(fs2.data.json.render.compact)
+        .compile
+        .drain)
+
+  @Benchmark
+  def intArrayPretty(bh: Blackhole) =
+    bh.consume(
+      intArrayStream
+        .through(fs2.data.json.render.prettyPrint())
+        .compile
+        .drain)
+
+  @Benchmark
+  def objectPretty(bh: Blackhole) =
+    bh.consume(
+      objectStream
+        .through(fs2.data.json.render.prettyPrint())
+        .compile
+        .drain)
+
+  @Benchmark
+  def intArrayPrettyLegacy(bh: Blackhole) =
+    bh.consume(
+      intArrayStream
+        .through(fs2.data.json.render.pretty())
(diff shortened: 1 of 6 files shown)


satabin:
@recons This PR should solve your problem with compact rendering once merged and released.

benchmarks/src/main/scala/fs2/data/benchmarks/PrinterBenchmarks.scala
```diff
@@ -0,0 +1,85 @@
...
+@Warmup(iterations = 15, time = 5)
+@Measurement(iterations = 10, time = 5)
+class PrinterBenchmarks {
+
+  val intArrayStream =
+    Stream.emits(
+      Token.StartArray ::
+        (List
+          .range(0, 1000000)
+          .map(i => Token.NumberValue(i.toString())) :+ Token.EndArray))
```
ybasket, line 28:
```suggestion
        (List.tabulate(1000000)(i => Token.NumberValue(i.toString())) :+ Token.EndArray))
```

Minor, but IMHO, that makes the intent a bit clearer.

I also would love to avoid appending to that long list, but I assume you want a single chunk? Otherwise using Stream's `++`, this could be simplified further.

text/shared/src/main/scala/fs2/data/text/render/internal/StreamPrinter.scala
```diff
@@ -21,246 +21,275 @@ import cats.collections.Dequeue
...
-  private def push(groups: Dequeue[(Int, Int, Chain[Annotated])],
-                   evt: Annotated): Dequeue[(Int, Int, Chain[Annotated])] =
-    groups.unsnoc match {
-      case Some(((ghpl, gindent, group), groups)) => groups.snoc((ghpl, gindent, group.append(evt)))
-      case None                                   => Dequeue.empty // should never happen
+  private def push(annctx: AnnotationContext, evt: Annotated): Unit =
+    annctx.groups.unsnoc match {
+      case Some((OpenGroup(ghpl, gindent, group), groups)) =>
+        annctx.groups = groups.snoc(OpenGroup(ghpl, gindent, group.append(evt)))
+      case None => // should never happen
```
ybasket, line 39: Do we have a better option than silently ignoring this bug? fs2 itself uses `assert()` to catch some bugs – not cool because AFAICT here we would still produce a semi-valid result, but let's at least have the discussion (and maybe a decision for all of fs2-data).

text/shared/src/main/scala/fs2/data/text/render/internal/StreamPrinter.scala
```diff
@@ -21,246 +21,275 @@ import cats.collections.Dequeue
...
+      ctx.fit += 1
+    }
+
+  private def renderGroupEnd(ctx: RenderingContext): Unit =
+    if (ctx.fit > 0) {
+      ctx.fit -= 1
+    }
+
+  private def renderIndentBegin(ctx: RenderingContext): Unit = {
+    ctx.lines = NonEmptyList(ctx.lines.head + (" " * indentSize), ctx.lines.tail)
```
ybasket, line 254: Do you think a `StringBuilder` that is allocated to the right capacity could help speeding up here? Like first append the head, then `indentSize` spaces?

ybasket (commented):
Looks good, just few minor comments.

benchmarks/src/main/scala/fs2/data/benchmarks/PrinterBenchmarks.scala
satabin: Yeah I want a single chunk, to see the overhead in processing one chunk.

text/shared/src/main/scala/fs2/data/text/render/internal/StreamPrinter.scala
satabin: I was think it would actually be way better to cache the padding. In a structured data rendering the same indent size will be found over and over again. I will try to reuse the padding that was already encountered for a indent depth, rather than this.

text/shared/src/main/scala/fs2/data/text/render/internal/StreamPrinter.scala
satabin: Yeah, I don't like cases like this one, where we cannot statically ensure never happen. And I don't like crashing with an `assert` either. I will try to find something better here. But it was already like that before 😬

text/shared/src/main/scala/fs2/data/text/render/internal/StreamPrinter.scala
satabin: I don't have much time right now to fine tune this. I will merge as is, since this fixes a big regression and publish a new bugfix release.

Issue #666, opened 2025-03-11
XML streaming hangs when using `fs2.data.xml.xpath.filter.raw`


mrdziuban:
I might be misunderstanding how to use `fs2.data.xml.xpath.filter.raw`, but my first attempt led to my code hanging forever.

Using a simple XML file, `test.xml`:

```xml
<?xml version='1.0' encoding='UTF-8'?>
<data>
  <test>text</test>
</data>
```

and running this code with `scala-cli /path/to/test.scala`:

```scala
//> using scala 3.6.4
//> using dep co.fs2::fs2-core:3.11.0
//> using dep co.fs2::fs2-io:3.11.0
//> using dep org.gnieh::fs2-data-xml:1.11.2

import cats.effect.{IO, IOApp}

object Test extends IOApp.Simple {
  def run: IO[Unit] =
    fs2.data.xml.xpath.XPathParser[IO]("/data/test").flatMap(xpath =>
      fs2.io.file.Files[IO].readAll(fs2.io.file.Path("/path/to/test.xml"))
        .through(fs2.text.utf8.decode)
        .through(fs2.data.xml.events[IO, String]())
        .through(fs2.data.xml.xpath.filter.raw(xpath))
        // This `debug` call prints `Stream(..)`
        .debug()
        // Nothing is ever printed by this `debug` call
        .evalMap(_.debug().compile.drain)
        .compile
        .drain
    )
}
```

The first `.debug()` call on the main `Stream` prints `Stream(..)`, but the `.debug().compile.drain` calls on the inner `Stream` don't seem to run, and the code never terminates.

mrdziuban:
Interestingly if I change `evalMap` to `parEvalMapUnbounded` then it works

satabin:
Hi. Thanks for reporting. I will try it locally and let you know.

satabin:
Hey I think I understand what the problem is.
With `evalMap` when the outer stream is pulled, on first matching element an inner stream is created. And this inner stream is compiled and drained, which will only return once the inner stream is closed. The inner stream is closed, when the match region is over. This means that the inner stream is waiting for more elements, which are pulled from the outer stream.

However, the outer stream is not pulled until the inner stream is finished, since `evalMap` will not pull until the effect has returned. So you have a stream that hangs.

[...]

PR #667, opened 2025-03-11
Add safer path operators and improve documentation

satabin:
The documentation now does not bring the unsafe `raw` operators first, and explains better what are the existing operators and what to be careful about.

This also adds a new `through` operator, that allows for streaming handling of path matches.

This PR is a follow-up on the discussion from #666 (cc @mrdziuban)


ybasket (approved):
Changes look good. 

I'm mildly unhappy with the naming of `through` as fs2 already uses that name for something similar, yet not equal and you get `through(filter.through(...))`. But I have no convincing alternative suggestion, `toPipe` and `funnel` both aren't really better. So if you have another good name, let's change it, otherwise merge as-is.

satabin:
> Changes look good.
> 
> I'm mildly unhappy with the naming of `through` as fs2 already uses that name for something similar, yet not equal and you get `through(filter.through(...))`. But I have no convincing alternative suggestion, `toPipe` and `funnel` both aren't really better. So if you have another good name, let's change it, otherwise merge as-is.

I kind of chose `through` on purpose, as it sends things to a `Pipe`, and this is how fs2 calls the action. I will merge and see if I can find something else before releasing.

PR #671, opened 2025-04-16
Add high-level Msgpack decoding API

jarmuszz:
This PR adds the high-level MessagePack decoding API.

The high-level API is split into two modules:
- `high.dynamic` which deserializes  data into wrapper objects, similarly to how `cbor.high` works.
- `high.static` which allows for deserialization into scala's native data-types.

The static module performs better than dynamic in terms of speed but is far less flexible, e.g. it currently lacks support for heterogeneous data.

The core of the implementation resides in the `MsgpackDecoder[A]` typeclass which represents a process of extracting values of type `A` from a `MsgpackItem` stream. Decoders can be ran via API endpoints specified in the package file. Instances of `MsgpackDecoder` for common built-in types are available via the `high.static.*` import. The `high.dynamic` module exports an instance of the typeclass for `MsgpackValue` and some methods to reduce boilerplate.

`MsgpackDecoder` is also monadic:

```scala
case class Record(id: Long, description: String, created: Instant)

implicit val recordDecoder: MsgpackDecoder[Record] =
  for {
    id      <- decoder[Long]
    desc    <- decoder[String]
    created <- decoder[Instant]
  } yield Record(id, desc, created)

val stream: Stream[IO, Byte] = ???
val records: IO[List[Records]] =
  stream
    .through(decode[Record])
    .compile
    .toList
```

I have also decided to ditch the old benchmark data because it consisted primarily of strings. The new bench file is a bit bigger (~32k vs ~27k MsgpackItems), hence the result difference for previous benchmarks.

```
MsgPackDecoderBenchmarks.dynamicDecoder         avgt   10  4256.090 ± 38.211  us/op
MsgPackDecoderBenchmarks.staticDecoder          avgt   10  3801.777 ± 36.186  us/op
MsgPackItemParserBenchmarks.parseMsgpackItems   avgt   10  4268.732 ± 58.199  us/op
MsgPackItemSerializerBenchmarks.serialize       avgt   10  4130.059 ±  9.869  us/op
MsgPackItemSerializerBenchmarks.withValidation  avgt   10  5653.779 ± 71.014  us/op
````

Benchmarks were done on i7-8550U.

I'm sure it would be possible to add typeclass derivation of the `MsgpackDecoder`. Sadly, I'm not really familiar with this part of Scala and I currently lack time to really dig into it.

I'm also not sure about the `static`/`dynamic` naming, would love some feedback :).

### CODE REVIEW

**Comment by jarmuszz** [comment]
Converting to draft as a couple of changes are needed to fix building for scalajs

**Comment by satabin** [comment]
> Converting to draft as a couple of changes are needed to fix building for scalajs

Is it now ready to be reviewed?

**Comment by jarmuszz** [comment]
Yes it is, thank you. I forgot to put up a comment.

**Review by satabin — COMMENTED** [review]
A had a first look and it looks generally ok, great work!!!!

I just am not sure either about the static/dynamic naming, and every time I looked at your PR I needed to re-read what you meant in the comment.

If we look at the JSON module, my understanding is that
 - `dynamic` is what we have with the `values` pipe in JSON. It produces a stream of JSON AST values out of the tokens (items in the msgpack implementation). It creates a structured representation of the input stream, which is a flat list of "events" describing the structure. What do you think about using the `values` name as well?
 - `static` is what we have with the `deserialize` pipe in JSON. It produces a stream of homogeneous typed values, out of the tokens (items in your case). What do you think about using the `deserialize` as well?

An other possibility for `values` could be `ast`, since this is what it produces. But I am not sure it helps with readability.

**Note:** In XML I used the DOM naming since this is a quite accepted term in this format. `values` would be `documents` in XML, since this is the name of the DOM root, and `deserialize` is called `elements` for XML.

Naming is hard :sweat_smile:

**Comment by jarmuszz** [comment]
I think it would be best to merge `high.static` into `high` and rename `dynamic` to `ast`.

- `high.static` -> `high`
- `high.dynamic` -> `high.ast`

Methods in `high` cannot be actually used without decoder instances which are provided via `static` or `dynamic` imports. While the former package relies on user calling `high`'s methods, the latter comes with it's own set of endpoints so it makes sense to keep it as a separate module. Merging `high.static` into `high` would simplify imports and make the package structure a bit cleaner.

As for the `dynamic`, renaming it into `values` could confuse users as there is a method with the same name inside the package. People would end-up calling `msgpack.high.values.values`. I think that `ast` works better in this case.

Also, changing wording from "decoding" to "deserialization" thought the code seems like a good idea to me.

Are these changes okay? If so, I'll push a commit in the following days.

Thank you for the feedback :)

**Comment by satabin** [comment]
> I think it would be best to merge `high.static` into `high` and rename `dynamic` to `ast`.
> 
>     * `high.static` -> `high`
> 
>     * `high.dynamic` -> `high.ast`
> 
> 
> Methods in `high` cannot be actually used without decoder instances which are provided via `static` or `dynamic` imports. While the former package relies on user calling `high`'s methods, the latter comes with it's own set of endpoints so it makes sense to keep it as a separate module. Merging `high.static` into `high` would simplify imports and make the package structure a bit cleaner.
> 
> As for the `dynamic`, renaming it into `values` could confuse users as there is a method with the same name inside the package. People would end-up calling `msgpack.high.values.values`. I think that `ast` works better in this case.
> 
> Also, changing wording from "decoding" to "deserialization" thought the code seems like a good idea to me.
> 
> Are these changes okay? If so, I'll push a commit in the following days.
> 
> Thank you for the feedback :)

This sounds reasonable, yes.

File: msgpack/shared/src/main/scala/fs2/data/msgpack/high/package.scala [hunk 1]
```diff
@@ -0,0 +1,72 @@
...
+    _.through(low.items[F]).through(fromItems[F, A])
+
+  /** Decodes a stream of bytes into a stream of `A`s via an explicit deserializer instance. 
+    */
+  @inline def deserialize[F[_], A](da: MsgpackDeserializer[A])(implicit F: RaiseThrowable[F]): Pipe[F, Byte, A] =
+    deserialize(F, da)
+
+  /** Summons a [[fs2.data.msgpack.high.MsgpackDeserializer]] instance for the desired type
+    */
+  def deserializer[A](implicit ev: MsgpackDeserializer[A]) = ev
```
satabin, line 71: A more idiomatic way of doing so is usually to have an `apply` method in the companion object of the typeclass. Like:

```scala
object MsgpackDeserializer {
  def apply[A](implicit ev: MsgpackDeserializer[A]): MsgpackDeserializer[A]=  ev
}
```

jarmuszz:
I have restructured the module as described above.

msgpack/shared/src/main/scala/fs2/data/msgpack/high/MsgpackDeserializer.scala
```diff
@@ -0,0 +1,76 @@
...
+import cats.Monad
+import fs2.data.msgpack.high.internal.Helpers._
+import scala.util.control.NonFatal
+
+/** Represents a process of extracting value of type `A` from an item stream.
+  *
+  * Built-in instances are accessbile via an import `fs2.data.msgpack.high.*`.
+  */
+trait MsgpackDeserializer[A] { outer =>
+  private[high] def run[F[_]: RaiseThrowable](ctx: DeserializationContext[F]): DeserializationResult[F, A]
```
satabin, line 31: I am not a fan of this signature. Writing a deserializer for a new type requires knowing a lot of the internals, if possible at all. Generally speaking I like deserializer that have zero knowledge of the context in which they are called, and that do not need to mutate anything (in you case you mutate the context). For me a deserializer should be a pure function from its inputs without any size effect, and with a public `deserialize` method, that can be tested and called without a big context.

Can you write it with this signature (or a slight variant thereof if it fits better)? That would not involved any fs2 specific thing.

```scala
trait MsgpackDeserializer[A] {
  def deserialize(bytes: ByteVector): DeserializationResult[A]
}

sealed trait DeserializationResult[+A]
object DeserializationResult {
  final case class Ok[A](value: A, reminder: ByteVector) extends DeserializationResult[A]
  final case class Error(msg: String) extends DeserializationResult[Nothing]
  final case class NeedsMoreBytes(hint: Option[Int]) extends DeserializationResult[Nothing]
}
```
The results models the three possible outcome:
 - We could deserialize and return the value and bytes that were not read
 - An error occurred with a message
 - We need more input to deserialize and want to be called again with more data (nothing was consumed) with a hint of how much we need

This is similar to the approach taken by [scodec](https://github.com/typelevel/fs2/blob/main/scodec/shared/src/main/scala/fs2/interop/scodec/StreamDecoder.scala) or by [framing](https://tokio.rs/tokio/tutorial/framing#buffered-reads) in tokio (hint is given by resizing the buffer). This way we decorrelate the decoding itself from acquiring the bytes, and make it easier to write custom deserializer (you do not need a `Pull` to decode)

What do you think about writing it this way?

satabin (changes requested):

msgpack/shared/src/main/scala/fs2/data/msgpack/high/MsgpackDeserializer.scala
jarmuszz: I think it should be possible to make the deserializer context-free the way you described. I have thought about design like this previously but I ended up with a conclusion that restarting the decoder over and over would be inefficient. Decoders would have to count the number of items in a buffer on every call which takes some time because we are working with structured data. Do you think there is a better way to handle this?

satabin: Correct me if I'm wrong, but msgpack encodes the length at the beginning for its compound types, right? If so, we can ensure that we read all the bytes before sending the data to decode to the decoder, limiting the roundtrips. What do you think?

jarmuszz: I rewrote the deserializer in [9922e91](https://github.com/gnieh/fs2-data/pull/671/commits/9922e91644d7d515699762619e17ebc55cb197d3). The code looks a lot cleaner now and it made it possible for the user to write their own deserializer instances.

Sadly, I was not able to achieve the same benchmark results as in the previous implementation. Although compound types encode the item count, we still have to account for nested data. Traversing through the item stream to check if we have `x` fully-buffered items always took significantly more time than just failing and restarting but this may not always be the case.

```
Benchmark                                        Mode  Cnt     Score    Error  Units
MsgpackDeserializerBenchmarks.deserialize        avgt   10  4524.216 ± 42.866  us/op
MsgpackDeserializerBenchmarks.deserializeValues  avgt   10  4668.547 ± 70.687  us/op
```

I have also encountered a problem with deserializer's `flatMap` calls pilling up on the stack. Looking through the fs2's `Pull` source code, I found a [`FlatMapOutput`](https://github.com/typelevel/fs2/blob/b27f4eda55933c5cbfa7d1bac93422be6472dd90/core/shared/src/main/scala/fs2/Pull.scala#L770) class and so I tried to implement [something similar](https://github.com/gnieh/fs2-data/pull/671/commits/9922e91644d7d515699762619e17ebc55cb197d3#diff-6784e229187a54e122abfc934b0d62d6786cd2d1e6236cad380130674f2ff044R54) for `MsgpackDeserializer`. This approach seems to solve the issue but I'm not sure if this is the best solution possible.

Let me know if you have any further comments and/or suggestions.

satabin: Thanks for the update. I have a week off coming so I will have time to look into it then.

msgpack/shared/src/main/scala/fs2/data/msgpack/high/package.scala
```diff
@@ -19,54 +19,65 @@ package data
...
+          uncons.flatMap {
+            case Some((head, tail)) => go(current ++ head.toVector, tail, false, None)
+            case None               => Pull.raiseError(new MsgpackUnexpectedEndOfStreamException())
+          }
+        } else if (current.isEmpty) {
+          Pull.output(Chunk.from(buffer.asScala)) >> rest.pull.uncons.flatMap {
+            case Some((head, tail)) =>
               buffer.clear()
-              go(DeserializationContext(hd, 0, tl), buffer)
+              go(current ++ head.toVector, tail, false, None)
```
satabin, line 52: Instead of converting to a `Vector` can't we use the `Chunk` type directly?

satabin: That's nice to read. Can you publish the `Vector` vs `Chunk` measurements? I am curious.

jarmuszz: Chunk:
```
MsgpackDeserializerBenchmarks.deserialize        avgt   10  2240.421 ± 41.723  us/op
MsgpackDeserializerBenchmarks.deserializeValues  avgt   10  2260.145 ± 69.666  us/op
```

Vector:
```
MsgpackDeserializerBenchmarks.deserialize        avgt   10  2959.630 ± 56.642  us/op
MsgpackDeserializerBenchmarks.deserializeValues  avgt   10  3010.632 ± 52.592  us/op
```


I'm on a new CPU so here are the other benchmarks to compare to:
```
MsgpackItemParserBenchmarks.parseMsgpackItems    avgt   10  2450.450 ± 35.014  us/op
MsgpackItemSerializerBenchmarks.serialize        avgt   10  2478.353 ± 40.145  us/op
MsgpackItemSerializerBenchmarks.withValidation   avgt   10  3092.503 ± 75.226  us/op
```

satabin (approved):

ybasket:
@jarmuszz Merging is blocked by the conflicts for CI. Do you have a chance to resolve them (merge `main`, then run `githubWorkflowGenerate`)?

PR #679, opened 2025-05-05
Make path operators more flexible

satabin:
This reworks the new XPath and JSONPath operators to make them more flexible:
 - the operator taking a `Pipe` that emits nothing is now called `consume` to reflect its purpose more closely
 - the `through` operator now takes any pipe, and will emit the output either deterministically or not depending on the `deterministic` parameter
 - JSONPath now also has a `filter.deserialize[Data](jsonpath)` pipe, that allows to deserialize selected data directly.


ybasket (approved):
LGTM!

Was slightly surprised that MiMa didn't complain, but well, better for us 😄

satabin:
Well, nothing I changed is released yet 😂

9 mai 2025 10:42:07 Yannick Heiber ***@***.***>:

PR #757, opened 2026-04-03
Introduce StaticHeaders type class for CSV

ybasket:
Currently, the only way to create an empty CSV with headers is by using a pipe that takes explicit headers. The more comfortable `encodeUsingFirstHeaders` requires the CSV to be non-empty. 

This PR adds a `encodeUsingStaticHeaders` pipe that takes the necessary header information from a new type class `StaticHeaders`. To facilitate adoption, the generic derivation now derives `CsvRowEncoder` instances that also implement `StaticHeaders`.

Still missing, will add once it's clear this PR is going somewhere:
* MiMa exclusions – it's unhappy about changes in `generic` because `DerivedCsvRowEncoder` is public. I don't think it would be a problem
* Documentation on the website (Scaladoc is already there)

### DIFF
--- a/csv/generic/shared/src/main/scala-2/fs2/data/csv/generic/internal/MapShapedCsvRowEncoder.scala
+++ b/csv/generic/shared/src/main/scala-2/fs2/data/csv/generic/internal/MapShapedCsvRowEncoder.scala
@@ -16,9 +16,8 @@
 
 package fs2.data.csv.generic.internal
 
-import cats.data.NonEmptyList
 import fs2.data.csv.generic.CsvName
-import fs2.data.csv.{CellEncoder, CsvRow, CsvRowEncoder}
+import fs2.data.csv.{CellEncoder, CsvRowEncoder}
 import shapeless._
 import shapeless.labelled._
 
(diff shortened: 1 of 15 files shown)


csv/generic/shared/src/main/scala-2/fs2/data/csv/generic/internal/MapShapedCsvRowEncoder.scala
```diff
@@ -30,26 +29,33 @@ object MapShapedCsvRowEncoder extends LowPrioMapShapedCsvRowEncoderImplicits {
...
+        Last(repr.head) :: Nil
+    }
 
 }
 
 private[generic] trait LowPrioMapShapedCsvRowEncoderImplicits {
   trait WithAnnotations[Repr, AnnoRepr] {
-    def fromWithAnnotation(row: Repr, annotation: AnnoRepr): CsvRow[String]
+    def headers(annotation: AnnoRepr): List[String]
+    def rawRow(repr: Repr): List[String]
```
ybasket, line 44: I'm using `List` to avoid allocating temporary `NonEmptyList` objects as we build up the columns. As it's purely internal and we're guaranteed to be non-empty, this should be 100% safe, but less memory-intensive.

csv/shared/src/main/scala/fs2/data/csv/RowF.scala
```diff
@@ -214,10 +214,10 @@ case class RowF[H[+a] <: Option[a], Header](values: NonEmptyList[String],
       order: Order[Header]): NonEmptyMap[Header, String] =
     headers.get.zip(values).toNem
 
-  /** Drop all headers (if any).
+  /** Drop all headers (if any). Retains the line number if set.
     * @return a row without headers, but same values
     */
-  def dropHeaders: Row = Row(values)
+  def dropHeaders: Row = Row(values, line)
```
ybasket, line 220: Noticed we were dropping line numbers silently here, fixed it along the way.

csv/shared/src/main/scala/fs2/data/csv/StaticHeaders.scala
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2024 fs2-data Project
```
satabin, line 2: I need to update that year...

csv/shared/src/main/scala/fs2/data/csv/StaticHeaders.scala
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package fs2.data.csv
+
+import cats.Functor
+import cats.data.NonEmptyList
+
+/** Type class for types that have statically known headers. This is useful for encoding/decoding CSV rows
+ * where the headers are fixed and can be determined at compile time. */
+trait StaticHeaders[T, H] {
```
satabin, line 24: I really think we need to drop this parametric header type for version 2, it makes things harder to read for no real benefit. Or do we have a strong point in favor of keeping it? I don't remember why I did that choice at the time...

ybasket: Yeah, I think its weight is not justified by its value. Especially as we also don't use it in the generic module (which could also be generic over the header type via the type classes instead of being hardcoded to `String`, but isn't). And something like a `NonEmptyMap[MyEnumType, String]` (roughly isomorphic to our `CsvRow`) can be decoded from `String` headers almost as easily.

satabin (approved):

PR #772, opened 2026-05-19
Add forColumns helpers to row-based CSV type classes

ybasket:
`(Csv)RowEncoder` and `(Csv)RowDecoder` companion objects now have forColumn helpers to create instances for product types easily. This aids manual creation of these type class instances significantly in circumstances where generic derivation is not desirable. 

Also extended the docs to mention the new helpers.

I checked the Java bytecode, it didn't explode in size and as expected, did not create any anonymous classes, just slim methods.

Questions for review:
1. Would you prefer this to live in `trait`s that are mixed in instead of having them directly on the companions?
2. If yes, would you prefer if the source code was generated rather than spelled out? circe does that, but I feel it only complicates the code.
3. Naming: Are you OK with `forColumns`? Note that I didn't add a number-based suffix like circe does. I don't think it is needed for the Scala versions we support and the use cases these methods see (you always need to provide types in some way) – kantan.csv has such same-name overloads and it seems to work fine + is more ergonomic imho.
4. You may notice missing tests – I don't think they would provide much value, but let me know if you disagree.


site/documentation/csv/index.md
```diff
@@ -314,6 +314,19 @@ val row = Stream(Option(3) :: "test" :: 42 :: HNil)
...
 ```
 
+For simple positional case classes, `RowDecoder.forColumns` and `RowEncoder.forColumns` build instances without boilerplate (overloads are provided for arities 1 through 22):
+
+```scala mdoc
+case class Triple(i: Int, s: String, j: Int)
+
+val tripleDecoder: RowDecoder[Triple] =
+  // Depending on your Scala version, you might also be able to use .forColumns(Triple.apply) instead
+  RowDecoder.forColumns((i: Int, s: String, j: Int) => Triple(i, s, j))
```
satabin: This shouldn't work as well, right? If so I'd prefer it this way in the doc.


```suggestion
  RowDecoder.forColumns(Triple(_, _, _))
```

site/documentation/csv/index.md
```diff
@@ -336,7 +349,22 @@ decoded.compile.toList
...
+As you can see this can be quite tedious to implement. The `CsvRowDecoder.forColumns` and `CsvRowEncoder.forColumns` helpers can save the boilerplate when you just want to map each named column to a constructor argument or a tuple element (overloads exist for arities 1 through 22). The encoder variant additionally returns a `StaticCsvRowEncoder`, exposing the headers via `StaticHeaders`:
+
+```scala mdoc
+case class NamedRow(i: Int, j: Int, s: String)
+
+val namedDecoder: CsvRowDecoder[NamedRow, String] =
+  // Depending on your Scala version, you might also be able to use .forColumns(NamedRow.apply) instead
+  CsvRowDecoder.forColumns("i", "j", "s") { (i: Int, j: Int, s: String) =>
+    NamedRow(i, j, s)
+  }
```
satabin: Same as above, I like this form better if it works:


```suggestion
  CsvRowDecoder.forColumns("i", "j", "s")(NamedRow(_, _, _))
```

site/documentation/csv/index.md
ybasket: Yes, that version should (usually) work, changed accordingly. The more ergonomic .apply variant does depend on the Scala version and even the type class (seems as enumerating the column names already resolves the overload in CsvRow(De|En)coder, so it works better there). Left those options in comment.

satabin (approved):

PR #776, opened 2026-06-01
Add ++ to concatenate CSV rows

ybasket:
When serializing nested case classes, it comes in handy to be able to concatenate CSV rows (with or without headers), hence this adds a `++` (feel free to bikeshed the name). Also adds `Order` instances for (CSV)Row as they're useful as well.

### DIFF
--- a/csv/shared/src/main/scala/fs2/data/csv/RowF.scala
+++ b/csv/shared/src/main/scala/fs2/data/csv/RowF.scala
@@ -219,6 +219,15 @@ case class RowF[H[+a] <: Option[a], Header](values: NonEmptyList[String],
     */
   def dropHeaders: Row = Row(values, line)
 
+  /**
+   * Combine two rows by concatenating their values and headers (if present). The resulting row will have no line number.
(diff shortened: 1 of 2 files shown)


csv/shared/src/main/scala/fs2/data/csv/RowF.scala
```diff
@@ -239,4 +248,8 @@ object RowF {
   implicit object functor extends Functor[CsvRow[*]] {
     override def map[A, B](fa: CsvRow[A])(f: A => B): CsvRow[B] = fa.copy(headers = Some(fa.headers.get.map(f)))
   }
+
+  implicit val orderRow: Order[Row] = Order.by(r => (r.values, r.line))
+  implicit def orderCsvRow[Header: Order]: Order[CsvRow[Header]] =
```
ybasket, line 253: I assume that for the usual header types an `Order` exists. If that's too strong, we can always introduce a lower priority implicit later.

satabin (approved):

PR #777, opened 2026-06-01
Use \r\n (CRLF) newlines by default

ybasket:
To fully comply with RFC 4180 by default, this changes the default newline to `\r\n`. It also extracts both the default separator and the newline values into constants so they're not repeated (potentially inconsistently) across the code.

### DIFF
--- a/csv/shared/src/main/scala/fs2/data/csv/package.scala
+++ b/csv/shared/src/main/scala/fs2/data/csv/package.scala
@@ -28,6 +28,16 @@ import scala.annotation.nowarn
 
 package object csv {
 
+  /**
+   * Default separator value for CSV data according to RFC-4180. Note that all pipes allow customizing this.
+   */
+  val csvDataDefaultSeparator: Char = ','
+
+  /**
+   * Default new line value for CSV data according to RFC-4180. Note that all pipes allow customizing this.
+   */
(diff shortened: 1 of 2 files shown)


csv/shared/src/main/scala/fs2/data/csv/package.scala
```diff
@@ -28,6 +28,16 @@ import scala.annotation.nowarn
 
 package object csv {
 
+  /**
+   * Default separator value for CSV data according to RFC-4180. Note that all pipes allow customizing this.
+   */
+  val csvDataDefaultSeparator: Char = ','
```
ybasket, line 34: I don't particularly like the `csvData` prefix, but people will usually have `import fs2.data.csv._` and I don't want source clashes with locally defined `defaultSeparator` and `defaultNewline` – just sitting on a codebase where those are actually defined 🙄 Open for better ideas.

satabin: Can't we make it something private? Ideally that would be an inlined const that never appears in the compiled files.

satabin:
Even though it is more spec compliant, doesn't this change mean that most of the people will now have to provide the line separator? It feels like a really annoying to identify breaking change to me.

PR #783, opened 2026-06-15
Make the Names given accessible to callers of deriveCsvRowDecoder/Enc…

### LINKED ISSUE #779: semiauto.deriveCsvRowDecoder summons private[generic] Names given
## Reproducer

```scala
//> using scala 3.8.4
//> using dep org.gnieh::fs2-data-csv-generic::1.14.0

package example

import fs2.data.csv.*
import fs2.data.csv.generic.semiauto.*

case class Foo(a: String, b: Int)

object Bar:
  given CsvRowDecoder[Foo, String] = deriveCsvRowDecoder
```

## Compiler output

```
  [info] -- Info: .../Bar.scala:9:50
  [info]  9 |  given CsvRowDecoder[Foo, String] = deriveCsvRowDecoder
  [info]    |                                                  ^
  [info]    |Usage of implicit given instance given_Names_T defined in object Names,
  [info]    |which is not accessible here. In Scala 3.10, this implicit will no longer be found.
```

(Triggered by [Implicits.warnIfImplicitFromInaccessibleCompanion](https://github.com/scala/scala3/blob/main/compiler/src/dotty/tools/dotc/typer/Implicits.scala) in Scala 3.)

Dichotomia:
…oder

`semiauto.deriveCsvRowDecoder` and `deriveCsvRowEncoder` are public and take a `using Names[T]`. That given is resolved at the *call site*, which lives outside `fs2.data.csv.generic`. Because `Names` and its given were `private[generic]`, the only way implicit search found them downstream was the deprecated inaccessible-companion fallback (Implicits.warnIfImplicitFromInaccessibleCompanion).

Since Scala 3.7 this emits a deprecation warning, and from Scala 3.10 the given will no longer be found at all, breaking derivation for every downstream user. Projects compiling with -Werror under 3.8.x are already blocked from upgrading.

Widen `Names` and its given to be publicly accessible so the given is found legitimately at the call site. No public signature changes, so this is binary compatible.

Adds a regression test that derives codecs from a foreign package; it reproduces the exact deprecation on the pre-fix code and compiles cleanly with the fix.

Fixes #779


csv/generic/shared/src/test/scala-3/fs2/data/csv/regression/InaccessibleNamesRegressionTest.scala
ybasket, line 1: Seems as the file path doesn't match the package (`generic` versus `regression`), could you please fix that?

ybasket (commented):
Thank you! Looks good to me, just one small fix still to do (if you find no time, I can eventually take care of it).

csv/generic/shared/src/test/scala-3/fs2/data/csv/regression/InaccessibleNamesRegressionTest.scala
Dichotomia: This test deliberately lives *outside* the `fs2.data.csv.generic` package to reproduce the call site of a downstream user. Do you want me to move it in an other directory ?

ybasket: No, sorry if I wasn't clear enough. It's good that the package is a different one, but the _file_ is still located in the `generic` folder, mismatching its package declaration. So it's just about moving the file to another folder – as said, I can take care of that if needed. Aligning on a release right now so this can go out soon.

Image

Dichotomia: Yes you can move the test to an other folder. Thanks for the help.

ybasket (approved):

